PR #15, opened 2020-07-09
Copied over files from opentelemetry-python repo

aabmass:
No actual newly written code here, this is all copied over from the opentelemetry-python repo. Also added a gitignore:
```bash
cp -R ../opentelemetry-python/ext/opentelemetry-exporter-cloud-* .
mkdir -p docs/examples && cp -R ../opentelemetry-python/docs/examples/cloud_* docs/examples/.
curl https://raw.githubusercontent.com/github/gitignore/master/Python.gitignore  > .gitignore
```

The next PR will have a basic tox set up, then I will work on CI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,138 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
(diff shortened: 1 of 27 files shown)


AndrewAXue:
There are also files in 

`opentelemetry-python/docs/examples/cloud_trace_exporter`
`opentelemetry-python/docs/examples/cloud_monitoring`

that we'll want as well.

aabmass:
> The GCP spec has a section "Directory structure" which we should be following here.
Discussed offline, will do in a another PR

I'll copy over the docs and set up building them in a separate PR

PR #17, opened 2020-07-10
Set up tox to run linters and tests

aabmass:
Copied over `.flake8`, `.isort.cfg`, `.pylintrc`, and `pyproject.toml` (black config) from opentelemetry-python and removed any references to directories not in this new repo.

I added a `tox.ini` file similar to the one in opentelemetry-python, but I changed a few things up:
- Added a tox factor `ci` for any envs that should run during CI. That way you can quickly run `tox -f ci -pauto` to see what would run in CI
- Added a constant `opentelemetry_python_commit` which is the commit hash in opentelemetry-python repo that CI should run against.
  * For now, the opentelemetry-python repo is changing rapidly between releases, and development here is concurrent with that
  * CI won't flake as opentelemetry-python changes
  * If you go through commit history, you can see what is the corresponding commit in opentelemetry-python repo
  * Once we hit RC or GA, we will remove this and just build against releases.
- This tox setup handles creating a development virtualenv for you, just run `tox -e dev` and it will make `venv/` in the repo root. It is the same virtualenv used for `tox -e lint` and `tox -e fix`, which should avoid any weird mismatches between these (I noticed this a lot in opentelemetry-python repo eachdist script).

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,15 @@
+[flake8]
+ignore =
+  E501  # line too long, defer to black
+  F401  # unused import, defer to pylint
+  W503  # allow line breaks before binary ops
+  W504  # allow line breaks after binary ops
+  E203  # allow whitespace before ':' (https://github.com/psf/black#slices) 
+exclude =
+  .git
+  .tox
+  CVS
+  .venv*/
+  venv*/
+  target
+  __pycache__
--- a/.gitignore
+++ b/.gitignore
@@ -136,3 +136,5 @@ dmypy.json
 
 # Cython debug symbols
 cython_debug/
+
+.vscode/
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,16 @@
+[settings]
+include_trailing_comma=True
+force_grid_wrap=0
+use_parentheses=True
+line_length=79
+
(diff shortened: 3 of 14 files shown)


aabmass:
Next dependent PR: https://github.com/aabmass/opentelemetry-operations-python/pull/1

AndrewAXue:
I don't think I'm experienced enough to know the possible repercussions of things like `opentelemetry_python_commit`. Pinging my managers to take a look.

sjbrunst:
I'm ok with the `opentelemetry_python_commit` approach; I agree that it would be good during the development phase to follow OpenTelemetry closer than the most recent release. Just a couple followups:

- Is there an issue created to remind us to remove it before GA?
- Is there something to remind/instruct contributors to bump the commit with their PRs?

aabmass:
> Is there an issue created to remind us to remove it before GA?

If everyone is ok with this approach, I can create one

> Is there something to remind/instruct contributors to bump the commit with their PRs?

I was planning to set up some automation with CircleCI to either file an issue (maybe a PR directly if I can figure it out) if it can be bumped, or file an issue if there is a breaking change in opentelemetry-python master.

@sjbrunst

sjbrunst:
Sounds good. LGTM!

PR #24, opened 2020-07-15
support GKE resource detection

AndrewAXue:
Add the ability to detect GKE resources in the Google Cloud Resource Detector and to export this info to Cloud Trace and Cloud Monitoring


opentelemetry-exporter-cloud-trace/src/opentelemetry/tools/resource_detector.py
```diff
@@ -42,5 +76,7 @@ def detect(self) -> "Resource":
             self.cached = True
             for resource_finder in _RESOURCE_FINDERS:
                 found_resources = resource_finder()
-                self.gcp_resources.update(found_resources)
+                if found_resources:
+                    self.gcp_resources = found_resources
+                    break
```
sjbrunst, line 81: If I understand this correctly, get_gke_resources() will run first. If it returns something, then we're done. If it returns nothing, then we'll try get_gce_resources() instead. And this only works because of the ordering of _RESOURCE_FINDERS.

This seems likely to break in future code changes, because it doesn't seem intuitive that the order of _RESOURCE_FINDERS should matter. Can you add a comment to _RESOURCE_FINDERS to explain why the order is important?

sjbrunst: Alternatively, I see that the extraction of fields like host.id and cloud.zone are duplicated in the two resource finders. Perhaps there's a way to restructure the code? Something like:

```
def get_google_resources()
  all_metadata = _get_all_google_metadata()
  resources = {}
  # add common resources to the map here
  if not container_name or not pod_namespace:
    # add gke specific resources to the map here
  return resources
```

opentelemetry-exporter-cloud-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
```diff
@@ -306,7 +306,16 @@ def _strip_characters(ot_version):
         "cloud.account.id": "project_id",
         "host.id": "instance_id",
         "cloud.zone": "zone",
-    }
+    },
+    "gke_container": {
```
sjbrunst, line 310: Where does this mapping get used? (It's not obvious from the rest of the CL why these lines are needed.)

opentelemetry-exporter-cloud-trace/src/opentelemetry/tools/resource_detector.py
```diff
@@ -28,7 +35,34 @@ def get_gce_resources():
...
+        "cloud.account.id": all_metadata["project"]["projectId"],
+        "k8s.cluster.name": all_metadata["instance"]["attributes"][
+            "cluster-name"
+        ],
+        "k8s.namespace.name": pod_namespace,
+        "host.id": all_metadata["instance"]["id"],
+        "k8s.pod.name": pod_name,
+        "container.name": container_name,
+        "cloud.zone": all_metadata["instance"]["zone"].split("/")[-1],
+        "cloud.provider": "gcp",
```
sjbrunst, line 59: Is there any chance that the resource detector gets called while running on a non-google machine? (The GKE and GCP detectors assume we're already on Google, I'm just making sure that's a valid assumption.)

opentelemetry-exporter-cloud-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
AndrewAXue: I generalized the conversion from OT resources -> MonitoredResource in the initial resources PR.
```
return {
        "g.co/r/{}/{}".format(resource_type, gcp_resource_key,): str(
            resource.labels[ot_resource_key]
        )
        for ot_resource_key, gcp_resource_key in OT_RESOURCE_LABEL_TO_GCP[
            resource_type
        ].items()
    }
```

opentelemetry-exporter-cloud-trace/src/opentelemetry/tools/resource_detector.py
AndrewAXue: I believe I discussed this with Chris and we agreed that if a user is calling the GoogleCloudResourceDetector they should be on a GCP tool. There is also an option to swallow that any errors of that nature by passing in `raise_on_error=False` to the detector.

opentelemetry-exporter-cloud-trace/src/opentelemetry/tools/resource_detector.py
AndrewAXue: I'd like to avoid making one large function as Cloud Monitoring might add support for more Google tools in the future, leading to one huge function. However, it does seem that `host.id` and `cloud.zone` are duplicated, I'll extract that out.

opentelemetry-exporter-cloud-trace/tests/test_gcp_resource_detector.py
```diff
@@ -46,11 +57,75 @@ def test_finding_gce_resources(self, getter):
...
+
+        pop_environ_key("CONTAINER_NAME")
+        pop_environ_key("NAMESPACE")
+        self.assertEqual(get_gke_resources(), {})
+
+        os.environ["NAMESPACE"] = "namespace"
+        self.assertEqual(get_gke_resources(), {})
+
+        pop_environ_key("NAMESPACE")
+        os.environ["CONTAINER_NAME"] = "container_name"
```
sjbrunst, line 90: Repeating a string like `"NAMESPACE"` or `"CONTAINER_NAME"` in code like this is vulnerable to typos; if it's mistyped in one spot it can break the test and be hard to debug.

A trick to avoid that is to define these strings in constants, then reuse that constant every time the magic string is needed. That will turn typos into compile errors (because the typo is in a variable name instead of a magic string) and make debugging those typos much easier.

opentelemetry-exporter-cloud-trace/tests/test_gcp_resource_detector.py
```diff
@@ -46,11 +57,75 @@ def test_finding_gce_resources(self, getter):
...
+        self.assertEqual(get_gke_resources(), {})
+        clear_gke_env_vars()
```
sjbrunst, line 92: this is called in `tearDown`; is it necessary here? (Same comment for other tests)

opentelemetry-exporter-cloud-trace/tests/test_gcp_resource_detector.py
```diff
@@ -46,11 +57,75 @@ def test_finding_gce_resources(self, getter):
...
+    pop_environ_key("NAMESPACE")
+    pop_environ_key("HOSTNAME")
+
+
+class TestGKEResourceFinder(unittest.TestCase):
+    def tearDown(self) -> None:
+        clear_gke_env_vars()
+
+    @mock.patch("opentelemetry.tools.resource_detector.requests.get")
+    def test_missing_environment_variables(self, getter):
```
sjbrunst, line 76: Let's break this down into even smaller tests:

```
def test_missing_container_name(self, getter):
  getter.return_value.json.return_value = GKE_RESOURCES_JSON_STRING
  pop_environ_key("CONTAINER_NAME")
  self.assertEqual(get_gke_resources(), {})

def test_missing_namespace(self, getter):
  getter.return_value.json.return_value = GKE_RESOURCES_JSON_STRING
  pop_environ_key("NAMESPACE")
  self.assertEqual(get_gke_resources(), {})
```

(The case where both are missing might be overkill so we can skip that. Then these tests are clear enough that I don't think you need the comment in lines 79-80.)

opentelemetry-exporter-cloud-trace/tests/test_gcp_resource_detector.py
```diff
@@ -46,11 +57,75 @@ def test_finding_gce_resources(self, getter):
...
+        getter.return_value.json.return_value = GKE_RESOURCES_JSON_STRING
```
sjbrunst, line 77: This line seems common to all the tests; is it possible to move it to a `setUp` method?

opentelemetry-exporter-cloud-trace/tests/test_gcp_resource_detector.py
AndrewAXue: The reason why I do this is do ensure a clean state between tests. Otherwise the order of the tests would matter.

opentelemetry-exporter-cloud-trace/tests/test_gcp_resource_detector.py
AndrewAXue: I don't believe we can move it to setUp, but we can add the decorater to the class instead of each individual test function. (which I've done)

aabmass:
@nilebox could you take a look at the kubernetes specific stuff here?

opentelemetry-exporter-cloud-trace/src/opentelemetry/tools/resource_detector.py
```diff
@@ -8,27 +10,71 @@
...
+        all_metadata,
+    ) = _get_google_metadata_and_common_attributes()
+    common_attributes.update(
+        {
+            "k8s.cluster.name": all_metadata["instance"]["attributes"][
+                "cluster-name"
+            ],
+            "k8s.namespace.name": pod_namespace,
+            "host.id": all_metadata["instance"]["id"],
+            "k8s.pod.name": os.getenv("HOSTNAME", ""),
```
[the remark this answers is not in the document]
nilebox: It's being set by Kubernetes I believe, but it doesn't always 100% match the pod name due to trimming: https://github.com/kubernetes/kubernetes/issues/4825#issuecomment-186058020

So I think it would be safer to read a separate environment variable by default, e.g. `POD_NAME`, and possibly fallback to `HOSTNAME` as a best guess.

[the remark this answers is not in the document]
nilebox: Though this behavior does match the OpenCensus implementation: https://github.com/census-ecosystem/opencensus-go-resource/blob/87ca538ec76b3961839940406845ef624d794fa8/gke/gke.go#L48

/cc @rghetia see the issue above about trimming the pod name to 63 characters. Shall we introduce a separate environment variable to be passed via https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-pod-fields-as-values-for-environment-variables ?

opentelemetry-exporter-cloud-trace/src/opentelemetry/tools/resource_detector.py
```diff
@@ -8,27 +10,71 @@
...
+    return common_attributes
+
+
+def get_gke_resources():
+    """ Resource finder for GKE attributes
+
+    """
+    # The user must specify these environment variables via the Downward API
+    container_name = os.getenv("CONTAINER_NAME")
+    pod_namespace = os.getenv("NAMESPACE")
```
nilebox, line 51: As discussed in the email thread, you can fallback to reading the `/var/run/secrets/kubernetes.io/serviceaccount/namespace` file if the `NAMESPACE` environment variable is undefined.

opentelemetry-exporter-cloud-trace/src/opentelemetry/tools/resource_detector.py
```diff
@@ -8,27 +10,71 @@
...
+
+        See: https://cloud.google.com/compute/docs/storing-retrieving-metadata
+    """
+    (
+        common_attributes,
+        all_metadata,
+    ) = _get_google_metadata_and_common_attributes()
+    common_attributes.update(
+        {
+            "host.id": all_metadata["instance"]["id"],
```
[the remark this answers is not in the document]
AndrewAXue: No, not all Google resources have a host.id (instance_id). e.g. https://cloud.google.com/monitoring/api/resources#tag_dataflow_job

opentelemetry-exporter-cloud-trace/tests/test_gcp_resource_detector.py
```diff
@@ -46,11 +61,65 @@ def test_finding_gce_resources(self, getter):
...
+        os.environ.pop(key)
+
+
+def clear_gke_env_vars():
+    pop_environ_key(CONTAINER_NAME)
+    pop_environ_key(NAMESPACE)
+    pop_environ_key(HOSTNAME)
+
+
+@mock.patch("opentelemetry.tools.resource_detector.requests.get")
```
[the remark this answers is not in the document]
AndrewAXue: Oh, very nice!

opentelemetry-exporter-cloud-trace/src/opentelemetry/tools/resource_detector.py
```diff
@@ -8,27 +10,71 @@
...
+
+
+def get_gke_resources():
+    """ Resource finder for GKE attributes
+
+    """
+    # The user must specify these environment variables via the Downward API
+    container_name = os.getenv("CONTAINER_NAME")
+    pod_namespace = os.getenv("NAMESPACE")
+    if not container_name or not pod_namespace:
```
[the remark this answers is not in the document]
AndrewAXue: An empty string for CONTAINER_NAME, seems like the user added the field but not a value. I guess we should still attempt GKE detection instead of silently falling back to GCE. Changed!

opentelemetry-exporter-cloud-trace/src/opentelemetry/tools/resource_detector.py
```diff
@@ -8,27 +10,71 @@
...
+        return {}
+    (
+        common_attributes,
+        all_metadata,
+    ) = _get_google_metadata_and_common_attributes()
+    common_attributes.update(
+        {
+            "k8s.cluster.name": all_metadata["instance"]["attributes"][
```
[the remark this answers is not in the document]
AndrewAXue: If a user was using k8s but not GKE, they shouldn't be using the GoogleCloudResourceDetector. Currently, all detectors are making the assumption that, if called, there IS something to detect. There is also the ability to pass in raise_on_error=False to swallow any errors.

opentelemetry-exporter-cloud-monitoring/src/opentelemetry/exporter/cloud_monitoring/__init__.py
```diff
@@ -26,7 +26,16 @@
...
-    }
+    },
+    "gke_container": {
+        "cloud.account.id": "project_id",
+        "k8s.cluster.name": "cluster_name",
+        "k8s.namespace.name": "namespace_id",
+        "host.id": "instance_id",
+        "k8s.pod.name": "pod_id",
+        "container.name": "container_name",
+        "cloud.zone": "zone",
```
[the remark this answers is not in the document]
AndrewAXue: On your second point, it seems python3.4 doesn't support that syntax. Reverting this change.

opentelemetry-exporter-cloud-trace/src/opentelemetry/tools/resource_detector.py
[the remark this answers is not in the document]
aabmass: I haven no idea how common this kind of setup is, but they still may want to detect GCE just not GKE.

opentelemetry-exporter-cloud-trace/src/opentelemetry/tools/resource_detector.py
```diff
@@ -8,27 +10,79 @@
...
+        all_metadata,
+    ) = _get_google_metadata_and_common_attributes()
+
+    # Fallback to reading namespace from a file is the env var is not set
+    pod_namespace = os.getenv("NAMESPACE")
+    if pod_namespace is None:
+        with open(
+            "/var/run/secrets/kubernetes.io/serviceaccount/namespace", "r"
+        ) as namespace_file:
+            pod_namespace = namespace_file.read().strip() or ""
```
aabmass, line 64: What is the `or ""` need for? If the file doesn't exist, you'd get an exception in the `open()` call, so if you want to handle that case you should do it explicitly

PR #29, opened 2020-07-17
Integrate mock server into trace integration test

the-ericwang35:
We'd like to integrate [our mock server](https://docs.google.com/document/d/1ZUwuqKIcTjrR32oxKwJw8XFNThQ8qDm_KvmTzzgiwtQ/edit#heading=h.h76dy5b7dqq2) into the Python exporter to be used for integration testing. 

To setup the server, Tox will curl the binaries from [our repository](https://github.com/googleinterns/cloud-operations-api-mock). Then, the tests themselves will spin up the server, run the integration tests, and teardown the server. 

Currently, the integration test just calls the mock server with the translated spans to verify that they follow the correct format and that the exporter has translated the spans properly. In the future, more tests can be added as needed

### DIFF
--- a/opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter_integration.py
+++ b/opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter_integration.py
@@ -1,0 +1,92 @@
+# Copyright The OpenTelemetry Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import socket
+import subprocess
+import unittest
+
+import grpc
+from google.cloud.trace_v2 import TraceServiceClient
+from google.cloud.trace_v2.gapic.transports import trace_service_grpc_transport
+from opentelemetry.exporter.cloud_trace import CloudTraceSpanExporter
+from opentelemetry.sdk.resources import Resource
+from opentelemetry.sdk.trace.export import Span, SpanExportResult
+from opentelemetry.trace import SpanContext, SpanKind
+
+
+class TestCloudTraceSpanExporter(unittest.TestCase):
+    @staticmethod
+    def get_free_port():
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sock.bind(("localhost", 0))
+        port = sock.getsockname()[1]
+        sock.close()
+        return port
+
+    def start_server(self):
+        port = self.get_free_port()
+
+        args = "./mock_server-x64-linux -address=localhost:" + str(port)
+        subprocess.Popen(args, shell=True)
+
+        return port
+
+    @staticmethod
+    def stop_server(port):
+        args = "fuser -k " + str(port) + "/tcp"
+        subprocess.Popen(args, shell=True)
+
+    def test_export(self):
+        port = self.start_server()
+
+        project_id = "TEST-PROJECT"
+        trace_id = "6e0c63257de34c92bf9efcd03927272e"
+        span_id = "95bb5edabd45950f"
+        resource_info = Resource(
+            {
+                "cloud.account.id": 123,
+                "host.id": "host",
+                "cloud.zone": "US",
+                "cloud.provider": "gcp",
+                "gcp.resource_type": "gce_instance",
+            }
+        )
+        span_data = [
+            Span(
+                name="span_name",
+                context=SpanContext(
+                    trace_id=int(trace_id, 16),
+                    span_id=int(span_id, 16),
+                    is_remote=False,
+                ),
+                parent=None,
+                kind=SpanKind.INTERNAL,
+                resource=resource_info,
+                attributes={"attr_key": "attr_value"},
+            )
+        ]
+
+        mock_server_address = "localhost:" + str(port)
+        channel = grpc.insecure_channel(mock_server_address)
+        transport = trace_service_grpc_transport.TraceServiceGrpcTransport(
+            channel=channel
+        )
+        client = TraceServiceClient(transport=transport)
+
+        exporter = CloudTraceSpanExporter(project_id, client=client)
+        result = exporter.export(span_data)
+
+        self.assertEqual(result, SpanExportResult.SUCCESS)
+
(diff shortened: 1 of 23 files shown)


opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter_integration.py
```diff
@@ -0,0 +1,82 @@
...
+class TestCloudTraceSpanExporter(unittest.TestCase):
+    def setUp(self):
+        # Find a free port to spin up our server at.
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sock.bind(("localhost", 0))
+        self.port = sock.getsockname()[1]
+        sock.close()
+
+        # Start our mock server.
+        args = "./mock_server-x64-linux -address=localhost:" + str(self.port)
```
aabmass, line 37: I think it might be a little cleaner to invoke with a non-relative path and assume the binary exists in `$PATH` (see also my below comment in tox.ini).

That way this code could be reused by users who want to write their own tests against the mock server for their projects.

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter_integration.py
```diff
@@ -0,0 +1,82 @@
...
+        subprocess.Popen(args, shell=True)
+
+    def tearDown(self):
+        args = "fuser -k " + str(self.port) + "/tcp"
+        subprocess.Popen(args, shell=True)
```
aabmass, line 42: You can save the output of L38's `Popen()` call and then just call [`kill()`](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.kill)

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter_integration.py
```diff
@@ -0,0 +1,82 @@
...
+import grpc
+from google.cloud.trace_v2 import TraceServiceClient
+from google.cloud.trace_v2.gapic.transports import trace_service_grpc_transport
+from opentelemetry.exporter.cloud_trace import CloudTraceSpanExporter
+from opentelemetry.sdk.resources import Resource
+from opentelemetry.sdk.trace.export import Span, SpanExportResult
+from opentelemetry.trace import SpanContext, SpanKind
+
+
+class TestCloudTraceSpanExporter(unittest.TestCase):
```
aabmass, line 28: Can you separate just the basic outline of this test (without any actual test cases) into a base class? That way all the integration tests that run against your mock can subclass that and not copy the setUp()/tearDown()

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter_integration.py
```diff
@@ -0,0 +1,82 @@
...
+                    is_remote=False,
+                ),
+                parent=None,
+                kind=SpanKind.INTERNAL,
+                resource=resource_info,
+                attributes={"attr_key": "attr_value"},
+            )
+        ]
+
+        mock_server_address = "localhost:" + str(self.port)
```
aabmass, line 72: nit this is used a few places, can we separate it into a property or similar?

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter_integration.py
```diff
@@ -0,0 +1,82 @@
...
+        channel = grpc.insecure_channel(mock_server_address)
+        transport = trace_service_grpc_transport.TraceServiceGrpcTransport(
+            channel=channel
+        )
+        client = TraceServiceClient(transport=transport)
+
+        exporter = CloudTraceSpanExporter(project_id, client=client)
+        result = exporter.export(span_data)
+
+        self.assertEqual(result, SpanExportResult.SUCCESS)
```
aabmass, line 82: Is it possible to check the written spans with a call to `ListTraces()` or is that in the works still?

tox.ini
```diff
@@ -67,6 +67,8 @@ changedir =
 
 commands_pre =
   test: pip install .
+  test: bash -c 'curl -O -L https://github.com/googleinterns/cloud-operations-api-mock/raw/master/cmd/mock_server-x64-linux'
+  test: bash -c 'chmod +x mock_server-x64-linux'
```
aabmass, line 71: I think a good choice would be to download this into [`{envbindir}`](https://tox.readthedocs.io/en/latest/config.html#substitutions-for-virtualenv-related-sections) which tox provides and already adds to `$PATH` when it runs tests.

Would be great to also check if the file is already downloaded and skip.

aabmass: I also think the `bash -c` is not needed here
```suggestion
  test: curl -O -L https://github.com/googleinterns/cloud-operations-api-mock/raw/master/cmd/mock_server-x64-linux
  test: chmod +x mock_server-x64-linux
```

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter_integration.py
the-ericwang35: Done

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter_integration.py
the-ericwang35: Done

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter_integration.py
the-ericwang35: Yep, done. Let me know if you want me to move it to another file or anything

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter_integration.py
the-ericwang35: Yep, done

tox.ini
the-ericwang35: Yep, changed to download into `{envbindir}` and added a check to avoid re-downloading the same file. For the `bash -c` part, if I remove it, it says stuff like `WARNING: test command found but not installed in testenv. cmd: /usr/bin/chmod. DEPRECATION WARNING: this will be an error in tox 4 and above!`

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter_integration.py
the-ericwang35: We do have that functionality, but we have to first compile the protos for that API to Python and add it to this repo since it's a custom endpoint we wrote. Is it okay if we do this in a separate PR?

the-ericwang35:
Messed up the rebase badly, just copied my changes to a clean branch and created a new PR here: https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/pull/37

PR #32, opened 2020-07-20
Document update

AndrewAXue:
Adding documentation and examples for the Cloud Trace Propagator and the Cloud resources detection features.


docs/examples/tools/cloud_resource_detector/resource_detector_metrics.py
```diff
@@ -0,0 +1,49 @@
...
+from opentelemetry.exporter.cloud_monitoring import (
+    CloudMonitoringMetricsExporter,
+)
+from opentelemetry.sdk.metrics import Counter, MeterProvider
+from opentelemetry.sdk.resources import get_aggregated_resources
+from opentelemetry.tools.resource_detector import GoogleCloudResourceDetector
+
+# MUST be run on a Google tool!
+# Detect resources from the environment
+resources = get_aggregated_resources([GoogleCloudResourceDetector()])
```
aabmass, line 28: Should this have `raise_on_error`?

docs/examples/tools/cloud_resource_detector/resource_detector_trace.py
```diff
@@ -0,0 +1,38 @@
...
+from opentelemetry import trace
+from opentelemetry.exporter.cloud_trace import CloudTraceSpanExporter
+from opentelemetry.sdk.resources import get_aggregated_resources
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import SimpleExportSpanProcessor
+from opentelemetry.tools.resource_detector import GoogleCloudResourceDetector
+
+# MUST be run on a Google tool!
+# Detect resources from the environment
+resources = get_aggregated_resources([GoogleCloudResourceDetector()])
```
aabmass, line 25: same comment

docs/examples/tools/cloud_resource_detector/resource_detector_metrics.py
AndrewAXue: I don't believe so, since its required that this code be run in a Google tool. The only case where `raise_on_error` should be used is code could be run both on AWS and GCE, meaning both GoogleCloudResourceDetector and AWSResourceDetector would be present (and obviously one failing). Otherwise, the user should just not attempt detection.

docs/examples/tools/cloud_resource_detector/resource_detector_trace.py
AndrewAXue: I don't believe so, since its required that this code be run in a Google tool. The only case where raise_on_error should be used is code could be run both on AWS and GCE, meaning both GoogleCloudResourceDetector and AWSResourceDetector would be present (and obviously one failing). Otherwise, the user should just not attempt detection.

opentelemetry-exporter-cloud-trace/src/opentelemetry/tools/resource_detector.py
```diff
@@ -106,4 +110,6 @@ def detect(self) -> "Resource":
                 if found_resources:
                     self.gcp_resources = found_resources
                     break
+        if not self.gcp_resources:
+            raise NoGoogleResourcesFound()
```
aabmass, line 114: From a debugging perspective, it would be great if we had a parent exception to `raise NoGoogleResourcesFound() from ex`. E.g. if the call to the metadata server timed-out, then the crash would show that.

AndrewAXue: resolved offline. We now log exceptions that were unexpected (network fail, keyerrors).

PR #34, opened 2020-07-21
Add config for readthedocs

aabmass:
Added `.readthedocs.yml` config to build sphinx docs, and also added a webhook for them to update on pushes and PRs. You can see the builds/docs, including those for PRs, here: https://readthedocs.org/projects/google-cloud-opentelemetry/builds/. Here is the built docs for this PR: https://google-cloud-opentelemetry--34.org.readthedocs.build/en/34/

Unfortunately I can't set up the PR status check which would add a link to each PR because without giving ReadTheDocs access to the GoogleCloudPlatform github org.

Last thing to note, I had to move the opentelemetry-python commit hash to `dev-constraints.txt` because ReadTheDocs won't let you specify a build command, you have to give it the dependencies and this was the easiest reusable way to do that. Unfortunately, because tox can't detect changes to external files like that, you may need to run `tox -r ...` to recreate virtualenvs when they becomes stale 😢

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -18,21 +18,10 @@ envlist =
 ; this section contains constants that can be referenced elsewhere
 [constants]
 
-; The commit in opentelemetry-python repo that the code in this repository is
-; built against. Use this for pre-GA development, and eventually we will build
-; against releases and remove this. For now, bump this frequently with every PR
-; to develop as closely with that repo as possible while still tracking the
-; actual hash.
-opentelemetry_python_commit = 545068d6eb61beaab031af45b1572d70e6652bc3
-opentelemetry_api_dep =
-  git+https://github.com/open-telemetry/opentelemetry-python.git@{[constants]opentelemetry_python_commit}#egg=opentelemetry-api&subdirectory=opentelemetry-api
-opentelemetry_sdk_dep =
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -18,21 +18,10 @@ envlist =
...
-opentelemetry_api_dep =
-  git+https://github.com/open-telemetry/opentelemetry-python.git@{[constants]opentelemetry_python_commit}#egg=opentelemetry-api&subdirectory=opentelemetry-api
-opentelemetry_sdk_dep =
-  git+https://github.com/open-telemetry/opentelemetry-python.git@{[constants]opentelemetry_python_commit}#egg=opentelemetry-sdk&subdirectory=opentelemetry-sdk
-
 base_deps =
   -c dev-constraints.txt
-  {[constants]opentelemetry_api_dep}
-  {[constants]opentelemetry_sdk_dep}
+  opentelemetry-api
```
AndrewAXue, line 23: Do these still need to be here if they're already in `dev-constraints.txt`?

aabmass: Yup, the constraint file (`-c` instead of `-r`) just specifies versions, not concrete dependencies. Pip just ignores lines in constraint files that aren't being installed

PR #35, opened 2020-07-22
Speed up Cloud Trace Exporter

AndrewAXue:
The previous flow was this. We would be passed a bunch of OT-spans. (Call these "ot-span")

Span(name='span_name', ...)

We would translate them in `_translate_to_cloud_trace` to a dictionary holding the contents of a Protobuf Span. (Call this a "translated span")

{
    'display_name': TruncatableString('span_name')
    ...
}

We would pass this to `CreateSpan` to get a proper Protobuf Span (Call this a "ProtoSpan")

ProtoSpan(display_name=TruncatableString('span_name'))

Finally we would pass all of these ProtoSpans to `batch_write_spans`. 
This would be extremely slow since we loop through each "translated span" to get the equivalent "ProtoSpan". `batch_write_spans` is quick in comparison because we pass it ALL of the "ProtoSpan"s at once. After some experimentation it turns out you can pass the list of "translated span"s to `batch_write_spans` directly, skipping the calls to `CreateSpan`. This PR implements that change. This speeds up BatchSpanProcessor by like 10x.


aabmass:
@AndrewAXue how did you measure this speed up and how slow was it before?

AndrewAXue:
> @AndrewAXue how did you measure this speed up and how slow was it before?

I created + exported 100 spans back to back using BatchSpanProcessor. It took 30 seconds before this change and now takes takes 3.2 seconds. The code for benchmarking (stripped down) looks like this 

```
from datetime import datetime
start_time = datetime.now()
for i in range(100):
    with tracer.start_as_current_span('foo'):
        print("Hello world!")
span_processor.force_flush()
end_time = datetime.now()
print((end_time - start_time).total_seconds())
```

aabmass:
`CreateSpan()` was actually calling the Cloud Trace API it looks like. Eric mentioned this was double writing the spans as well, is that right?

aabmass:
@AndrewAXue, what does `CreateSpan()` do then? I'm just a little confused if the second call wasn't doing a second write, the docs say "you cannot update existing spans" and "the span names must not match existing spans, or the results are undefined."

https://cloud.google.com/trace/docs/reference/v2/rpc/google.devtools.cloudtrace.v2#batchwritespansrequest

docs/examples/cloud_trace_exporter/more_trace.py
```diff
@@ -0,0 +1,43 @@
+from opentelemetry import trace
```
aabmass, line 1: Do we want to include this example in the docs? If so, it will need an rst

AndrewAXue: This is from another PR whoops, removing.

PR #36, opened 2020-07-22
add examples of more trace exporter features

AndrewAXue:
This new example shows off some more things that Cloud Trace Exporter is capable of, like adding attributes to spans, adding events to spans, adding links, etc.

### DIFF
--- a/docs/examples/cloud_trace_exporter/README.rst
+++ b/docs/examples/cloud_trace_exporter/README.rst
@@ -20,7 +20,7 @@ To use this exporter you first need to:
     pip install opentelemetry-sdk
     pip install opentelemetry-exporter-cloud-trace
 
-* Run an basic example locally
+* Run a basic example locally
 
 .. literalinclude:: basic_trace.py
     :language: python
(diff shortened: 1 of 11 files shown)


docs/examples/cloud_trace_exporter/more_trace.py
```diff
@@ -0,0 +1,43 @@
...
+
+# Adding links to spans
+with tracer.start_as_current_span("link_target") as link_target:
+    with tracer.start_as_current_span(
+        "span_with_link", links=[Link(link_target.context)]
+    ):
+        pass
+    with tracer.start_as_current_span(
+        "span_with_link_and_attributes",
+        links=[Link(link_target.context, attributes={"link_attr": "string"})],
```
sjbrunst, line 34: Does this attribute get applied to `link_target` or the new span? `link_attr` sounds like it's an attribute of the linked trace (`link_target`), but `span_with_link_and_attributes` makes it sound like the new span is the one with attributes. Some code comments or renaming could help here.

AndrewAXue: The attribute actually gets applied to the Link itself. I'll add comments clarifying that.

docs/examples/cloud_trace_exporter/README.rst
```diff
@@ -20,12 +20,18 @@ To use this exporter you first need to:
     pip install opentelemetry-sdk
     pip install opentelemetry-exporter-cloud-trace
 
-* Run an example locally
+* Run an basic example locally
```
aabmass, line 23:
```suggestion
* Run a basic example locally
```

PR #37, opened 2020-07-23
Add integration tests for trace using mock server

the-ericwang35:
We'd like to integrate [our mock server](https://docs.google.com/document/d/1ZUwuqKIcTjrR32oxKwJw8XFNThQ8qDm_KvmTzzgiwtQ/edit#heading=h.h76dy5b7dqq2) into the Python exporter to be used for integration testing.

To setup the server, Tox will curl the binaries from [our repository](https://github.com/googleinterns/cloud-operations-api-mock). Then, the tests themselves will spin up the server, run the integration tests, and teardown the server.

Currently, the integration test just calls the mock server with the translated spans to verify that they follow the correct format and that the exporter has translated the spans properly. In the future, more tests can be added as needed

### DIFF
--- a/opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
+++ b/opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
@@ -1,0 +1,88 @@
+# Copyright The OpenTelemetry Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import socket
+import subprocess
+import unittest
+from time import sleep
+
+import grpc
+from google.cloud.trace_v2 import TraceServiceClient
+from google.cloud.trace_v2.gapic.transports import trace_service_grpc_transport
+from opentelemetry.exporter.cloud_trace import CloudTraceSpanExporter
+from opentelemetry.sdk.resources import Resource
+from opentelemetry.sdk.trace.export import Span, SpanExportResult
+from opentelemetry.trace import SpanContext, SpanKind
+
+
+class BaseExporterIntegrationTest(unittest.TestCase):
+    def setUp(self):
+        # Find a free port to spin up our server at.
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sock.bind(("localhost", 0))
+        self.address = "localhost:" + str(sock.getsockname()[1])
+        sock.close()
+
+        # Start the mock server.
+        args = "mock_server-x64-linux -address=" + self.address
+        self.mock_server_process = subprocess.Popen(args, shell=True)
+
+    def tearDown(self):
+        self.mock_server_process.kill()
+
+
+class TestCloudTraceSpanExporter(BaseExporterIntegrationTest):
+    def test_export(self):
+        project_id = "TEST-PROJECT"
+        trace_id = "6e0c63257de34c92bf9efcd03927272e"
+        span_id = "95bb5edabd45950f"
+
+        # Create span and associated data.
+        resource_info = Resource(
+            {
+                "cloud.account.id": 123,
+                "host.id": "host",
+                "cloud.zone": "US",
+                "cloud.provider": "gcp",
+                "gcp.resource_type": "gce_instance",
+            }
+        )
+        span = Span(
+            name="span_name",
+            context=SpanContext(
+                trace_id=int(trace_id, 16),
+                span_id=int(span_id, 16),
+                is_remote=False,
+            ),
+            parent=None,
+            kind=SpanKind.INTERNAL,
+            resource=resource_info,
+            attributes={"attr_key": "attr_value"},
+        )
+        span.start()
+        sleep(0.1)
+        span.end()
+        span_data = [span]
+
+        # Setup the exporter.
+        channel = grpc.insecure_channel(self.address)
+        transport = trace_service_grpc_transport.TraceServiceGrpcTransport(
+            channel=channel
+        )
+        client = TraceServiceClient(transport=transport)
+        exporter = CloudTraceSpanExporter(project_id, client=client)
+
+        # Export the spans and verify the results.
+        result = exporter.export(span_data)
+        self.assertEqual(result, SpanExportResult.SUCCESS)
--- a/opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
+++ b/opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
@@ -15,7 +15,6 @@
 import socket
 import subprocess
 import unittest
-from time import sleep
 
 import grpc
 from google.cloud.trace_v2 import TraceServiceClient
@@ -24,10 +23,13 @@ from opentelemetry.exporter.cloud_trace import CloudTraceSpanExporter
 from opentelemetry.sdk.resources import Resource
 from opentelemetry.sdk.trace.export import Span, SpanExportResult
 from opentelemetry.trace import SpanContext, SpanKind
+from opentelemetry.util import time_ns
 
 
 class BaseExporterIntegrationTest(unittest.TestCase):
     def setUp(self):
+        self.project_id = "TEST-PROJECT"
+
         # Find a free port to spin up our server at.
         sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         sock.bind(("localhost", 0))
@@ -35,8 +37,12 @@ class BaseExporterIntegrationTest(unittest.TestCase):
         sock.close()
 
         # Start the mock server.
-        args = "mock_server-x64-linux -address=" + self.address
-        self.mock_server_process = subprocess.Popen(args, shell=True)
+        args = ["mock_server-x64-linux", "-address", self.address]
+        self.mock_server_process = subprocess.Popen(
+            args, stderr=subprocess.PIPE
+        )
+        # Block until the mock server starts (it will output the address after starting).
+        self.mock_server_process.stderr.readline()
 
     def tearDown(self):
         self.mock_server_process.kill()
@@ -44,7 +50,6 @@ class BaseExporterIntegrationTest(unittest.TestCase):
 
 class TestCloudTraceSpanExporter(BaseExporterIntegrationTest):
     def test_export(self):
-        project_id = "TEST-PROJECT"
         trace_id = "6e0c63257de34c92bf9efcd03927272e"
         span_id = "95bb5edabd45950f"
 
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -67,6 +67,8 @@ changedir =
 
 commands_pre =
   test: pip install .
+  test: bash -c '[[ -e {envbindir}/mock_server-x64-linux ]] || curl -L -o {envbindir}/mock_server-x64-linux https://github.com/googleinterns/cloud-operations-api-mock/raw/master/cmd/mock_server-x64-linux'
```
the-ericwang35, line 70: @aabmass Regarding you previous comment about not needing the `bash -c`, it still works without it, however I get a bunch of warnings: `WARNING: test command found but not installed in testenv. cmd: /usr/bin/chmod. DEPRECATION WARNING: this will be an error in tox 4 and above!`

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
```diff
@@ -0,0 +1,88 @@
...
+        channel = grpc.insecure_channel(self.address)
+        transport = trace_service_grpc_transport.TraceServiceGrpcTransport(
+            channel=channel
+        )
+        client = TraceServiceClient(transport=transport)
+        exporter = CloudTraceSpanExporter(project_id, client=client)
+
+        # Export the spans and verify the results.
+        result = exporter.export(span_data)
+        self.assertEqual(result, SpanExportResult.SUCCESS)
```
the-ericwang35, line 88: @aabmass Regarding your previous comment about adding a call to `GetNumSpans` from our server, this would be possible, however we'd have to first compile the protos for that API to Python and add it to this repo to give access to the client, since it's a custom endpoint we wrote. Is it okay if we do this in a separate PR?

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
```diff
@@ -0,0 +1,88 @@
...
+                span_id=int(span_id, 16),
+                is_remote=False,
+            ),
+            parent=None,
+            kind=SpanKind.INTERNAL,
+            resource=resource_info,
+            attributes={"attr_key": "attr_value"},
+        )
+        span.start()
+        sleep(0.1)
```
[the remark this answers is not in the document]
AndrewAXue: Yeah we definately don't want to use `sleep` in any test

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
```diff
@@ -0,0 +1,88 @@
...
+        span.end()
+        span_data = [span]
```
AndrewAXue, line 76: Instead of doing `span.start()` `span.end()` can you just manually set the `end_time` and `start_time`

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
```diff
@@ -0,0 +1,88 @@
...
+        # Start the mock server.
+        args = "mock_server-x64-linux -address=" + self.address
+        self.mock_server_process = subprocess.Popen(args, shell=True)
+
+    def tearDown(self):
+        self.mock_server_process.kill()
+
+
+class TestCloudTraceSpanExporter(BaseExporterIntegrationTest):
+    def test_export(self):
```
AndrewAXue, line 46: Can we add some more tests for the various features that the Exporter supports? Events, Links, Status Codes, etc.

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
```diff
@@ -0,0 +1,88 @@
...
+        project_id = "TEST-PROJECT"
```
AndrewAXue, line 47: Not sure if this makes sense to have, but can we have a test where we purposefully try to export a malformed span and assert that the export is failing?

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
```diff
@@ -0,0 +1,88 @@
...
+        span.end()
+        span_data = [span]
+
+        # Setup the exporter.
+        channel = grpc.insecure_channel(self.address)
+        transport = trace_service_grpc_transport.TraceServiceGrpcTransport(
+            channel=channel
+        )
+        client = TraceServiceClient(transport=transport)
+        exporter = CloudTraceSpanExporter(project_id, client=client)
```
AndrewAXue, line 84: This is super clean nice!

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
AndrewAXue, line 76: Can we have tests where we export more then one spans?

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
```diff
@@ -0,0 +1,88 @@
...
+    def setUp(self):
+        # Find a free port to spin up our server at.
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sock.bind(("localhost", 0))
+        self.address = "localhost:" + str(sock.getsockname()[1])
+        sock.close()
+
+        # Start the mock server.
+        args = "mock_server-x64-linux -address=" + self.address
+        self.mock_server_process = subprocess.Popen(args, shell=True)
```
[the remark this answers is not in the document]
the-ericwang35: Good catch, removed

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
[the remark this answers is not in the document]
the-ericwang35: Removed the `sleep`

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
the-ericwang35: Yep, done

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
the-ericwang35: I was definitely planning on adding more tests but in a separate PR, the main purpose of this PR was to just get the pipeline/base integration test class setup so that Mohamed can start writing the metric tests in parallel. I'll write some tests for all of the things you mentioned in the next PR.

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
the-ericwang35: Yep same as above, I'll add this in a separate PR

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
the-ericwang35: Yep same as above, I'll add this in a separate PR

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
the-ericwang35: As discuessed, `ListTraces()` is from cloud trace v1, and we're using cloud trace v2.

tox.ini
```diff
@@ -67,6 +67,8 @@ changedir =
...
+  test: bash -c 'chmod +x {envbindir}/mock_server-x64-linux'
```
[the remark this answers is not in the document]
the-ericwang35: Done, moved it to a `.sh` file

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
AndrewAXue: Gotcha, can you make an issue in the repo tracking this?

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
AndrewAXue: As above, can you make an issue in the repo tracking this?

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
AndrewAXue: As above, can you make an issue in the repo tracking this?

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
the-ericwang35: Yet, created one here: https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/issues/41 (it won't let me assign it to myself, so feel free to assign it to me)

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
AndrewAXue: You should also do `span.end(...)` in that case. 

I'm still concerned about this approach since `.start()` `.end()` do a lot of other stuff like interacting with the `span_processor`. Currently this is a no-op, but in the future this may change (e.g. SpanProcessor becomes an abstract class the throws if it gets initialized + called). I still think we should set `_start_time` `_end_time` manually and just tell pylint to not complain. @aabmass any thoughts?

the-ericwang35: Gotcha, if `.start()` do unwanted things then I agree that setting the fields would be better, so I've made this change.

aabmass: If you want to completely get rid of any side effects from `Span`, you can use a `Mock()`. You might be able to just replace your call to `Span()` above with `Mock()` (same exact arguments) e.g.

```py
span = Mock(
    spec=Span,
    start_time=whenever,
    end_time=whenver + 10 ** 9,

    name="span_name",
    context=SpanContext(
        trace_id=int(trace_id, 16), span_id=int(span_id, 16), is_remote=False,
    ),
    parent=None,
    kind=SpanKind.INTERNAL,
    resource=resource_info,
    attributes={"attr_key": "attr_value"},
)
```

The `spec=Span` kwarg to Mock  makes it raise an error if the Span is misused also

tox.ini
```diff
@@ -67,6 +67,8 @@ changedir =
 
 commands_pre =
   test: pip install .
+  test: bash -c 'chmod +x {toxinidir}/get_mock_server.sh'
```
aabmass, line 70: I don't think this is needed

opentelemetry-exporter-cloud-trace/tests/test_integration_cloud_trace_exporter.py
the-ericwang35: I tried using `mock.Mock()` and it gives `return "{:032x}".format(trace_id)` `TypeError: unsupported format string passed to Mock.__format__` :(

tox.ini
the-ericwang35: Yep, removed

PR #38, opened 2020-07-23
Test none link attributes

AndrewAXue:
I previously forgot to deal with the case where a Link has "None" as its attributes. The issue was (accidentally) merged in another PR (https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/pull/35), and this PR adds a unit test for it.


opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter.py
```diff
@@ -306,6 +306,30 @@ def test_extract_links(self):
...
 
+    def test_extract_link_with_none_attribute(self):
+        trace_id = "6e0c63257de34c92bf9efcd03927272e"
+        span_id = "95bb5edabd45950f"
+        link = Link(
+            context=SpanContext(
+                trace_id=int(trace_id, 16),
+                span_id=int(span_id, 16),
+                is_remote=False,
+            ),
```
aabmass, line 317: nit, for clarity since it took me a second to realize what was happening here

```suggestion
            ),
            attributes=None,
```

PR #40, opened 2020-07-23
refactor cloud trace tests

AndrewAXue:
Refactoring Cloud Trace tests according to feedback. The goal of this PR was to  split up tests into more "bite-size" pieces where the purpose would be immediately obvious from the name, and adding more clarifying comments where needed.


opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter.py
```diff
@@ -182,28 +185,102 @@ def test_extract_status(self):
...
+                attribute_map={"legal_attribute": AttributeValue(int_value=3)},
+                dropped_attributes_count=1,
+            ),
         )
 
+    def test_too_many_attributes(self):
         too_many_attrs = {}
         for attr_key in range(5):
             too_many_attrs[str(attr_key)] = 0
         proto_attrs = _extract_attributes(too_many_attrs, 4)
```
sjbrunst, line 214: To make this function a little more self-documenting you could write it as:

```
def test_too_many_attributes(self):
    attr_limit = 4
    too_many_attrs = {}
    for attr_key in range(attr_limit + 1):
        too_many_attrs[str(attr_key)] = 0
    proto_attrs = _extract_attributes(too_many_attrs, attr_limit)
    self.assertEqual(proto_attrs.dropped_attributes_count, 1)
```

Alternatively, using `num_attrs_limit=4` instead of just `4` as mentioned in another comment would be enough to improve the readability of this test.

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter.py
```diff
@@ -182,28 +185,102 @@ def test_extract_status(self):
...
 
-    def test_extract_attributes(self):
+    def test_extract_empty_attributes(self):
         self.assertEqual(
             _extract_attributes({}, 4), ProtoSpan.Attributes(attribute_map={})
         )
+
+    def test_extract_variety_of_attributes(self):
         self.assertEqual(
             _extract_attributes(self.attributes_variety_pack, 4),
```
sjbrunst, line 195: (this comment applies to all tests): Generally speaking you shouldn't assume that the reader of the test (or any code) knows what all the parameters are to the function you're calling. In this case `self.attributes_variety_pack` is a descriptive name, so it's clear to the reader that the first argument to `_extract_attributes` is some sort of collection of attributes. But the reader would have no idea what the `4` means nor whether it's important to the test unless they go and look up the definition of `_extract_attributes`.

There are two main ways to improve this: put the 4 in a variable/constant with a descriptive name, or (since Python allows you to name your parameters) simply replace the `4` with `num_attrs_limit=4`.

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter.py
```diff
@@ -182,28 +185,102 @@ def test_extract_status(self):
...
-    def test_extract_events(self):
+    def test_add_agent_attribute(self):
+        self.assertEqual(
+            _extract_attributes({}, 4, add_agent_attr=True),
+            ProtoSpan.Attributes(
+                attribute_map={"g.co/agent": self.agent_code},
+                dropped_attributes_count=0,
+            ),
+        )
+        # Drop existing attributes in favor of the agent attribute
```
sjbrunst, line 225: Should this be broken into two tests?

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter.py
```diff
@@ -182,28 +185,102 @@ def test_extract_status(self):
...
+                dropped_attributes_count=1,
+            ),
+        )
+
+    def test_extract_empty_events(self):
         self.assertIsNone(_extract_events([]))
+
+    def test_too_many_events(self):
+        time_in_ns = 1589919268850900051
+        time_in_ms_and_ns = {"seconds": 1589919268, "nanos": 850899968}
```
sjbrunst, line 241: Should these numbers match? `time_in_ns` ends with `900051` but `time_in_ms_and_ns["nanos"]` ends with `899968`.

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter.py
```diff
@@ -182,28 +185,102 @@ def test_extract_status(self):
...
+                    },
+                ]
+                * MAX_NUM_EVENTS,
+                dropped_annotations_count=len(too_many_events)
+                - MAX_NUM_EVENTS,
+            ),
+        )
+
+    def test_too_many_event_attributes(self):
+        time_in_ns1 = 1589919268850900051
```
sjbrunst, line 263: Since there's no `time_in_ns2`, you can just call this `time_in_ns`.

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter.py
```diff
@@ -182,28 +185,102 @@ def test_extract_status(self):
...
+        self.assertEqual(
+            proto_events.time_event[
+                0
+            ].annotation.attributes.dropped_attributes_count,
+            len(event_attrs) - MAX_EVENT_ATTRS,
+        )
+
+    def test_extract_multiple_events(self):
         time_in_ns1 = 1589919268850900051
         time_in_ms_and_ns1 = {"seconds": 1589919268, "nanos": 850899968}
```
sjbrunst, line 285: Same comment here about numbers not matching.

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter.py
```diff
@@ -246,30 +323,32 @@ def test_extract_events(self):
             ),
         )
 
-    def test_extract_links(self):
+    def test_extract_empty_links(self):
         self.assertIsNone(_extract_links([]))
-        trace_id = "6e0c63257de34c92bf9efcd03927272e"
+
+    def test_extract_multiple_links(self):
+        self.example_trace_id = "6e0c63257de34c92bf9efcd03927272e"
```
sjbrunst, line 330: example_trace_id is defined in `setUp()`, is this needed?

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter.py
AndrewAXue: Done! I've also made some other stuff into named arguments like the truncate_str args

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter.py
AndrewAXue: Seems we're run into a bit of numerical uncertainty

sjbrunst: This is addressed in PR #42.

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter.py
AndrewAXue: This is addressed in PR #42.

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter.py
```diff
@@ -76,6 +77,23 @@ def setUp(self):
...
+        )
+        self.example_trace_id = "6e0c63257de34c92bf9efcd03927272e"
+        self.example_span_id = "95bb5edabd45950f"
+        self.example_time_in_ns = 1589919268850900051
+        self.example_time_stamp = Timestamp(
+            seconds=1589919268, nanos=850900051
+        )
+        self.str_300 = "a" * 300
+        self.str_256 = "a" * 256
+        self.str_128 = "a" * 128
```
aabmass, line 96: If these are just immutable constants that don't need to be re-initialized for each test run, they can just be class constants IMO. Same goes for any of the above that can easily be expressed as class constants

AndrewAXue: Resolved offline

opentelemetry-exporter-cloud-monitoring/tests/test_cloud_monitoring.py
```diff
@@ -70,16 +70,18 @@ def __init__(
 
 
 class TestCloudMonitoringMetricsExporter(unittest.TestCase):
-    def setUp(self):
-        self.client_patcher = mock.patch(
+    @classmethod
+    def setUpClass(cls):
+        cls.client_patcher = mock.patch(
```
aabmass, line 75: This patch might be better off in `setUp()` or iirc you can decorate the class with it. It doesn't look like you're using the mock in any of the test cases, but it would accumulate state (e.g. `mock.method_calls`) across test cases.

My original comment was regarding immutable constants

opentelemetry-exporter-cloud-trace/tests/test_cloud_trace_exporter.py
```diff
@@ -47,20 +47,22 @@
 from opentelemetry.trace.status import StatusCanonicalCode
 
 
+# pylint: disable=too-many-public-methods
 class TestCloudTraceSpanExporter(unittest.TestCase):
-    def setUp(self):
-        self.client_patcher = mock.patch(
+    @classmethod
+    def setUpClass(cls):
+        cls.client_patcher = mock.patch(
```
aabmass, line 54: same comment as above.

PR #42, opened 2020-07-24
fix floating point error

AndrewAXue:
Fixing a floating point error caused by using 1e9 instead of 10**9

### DIFF
--- a/opentelemetry-exporter-cloud-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
+++ b/opentelemetry-exporter-cloud-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
@@ -182,7 +182,7 @@ def _get_time_from_ns(nanoseconds: int) -> Dict:
     nanoseconds"""
     if not nanoseconds:
         return None
-    seconds, nanos = divmod(nanoseconds, 1e9)
+    seconds, nanos = divmod(nanoseconds, 10 ** 9)
     return {"seconds": int(seconds), "nanos": int(nanos)}
 
 
--- a/opentelemetry-exporter-cloud-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
+++ b/opentelemetry-exporter-cloud-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
@@ -178,7 +178,7 @@ class CloudTraceSpanExporter(SpanExporter):
         pass
 
 
-def _get_time_from_ns(nanoseconds: int) -> Dict:
+def _get_time_from_ns(nanoseconds: int) -> Optional[Timestamp]:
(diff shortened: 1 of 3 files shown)


opentelemetry-exporter-cloud-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
```diff
@@ -182,7 +182,7 @@ def _get_time_from_ns(nanoseconds: int) -> Dict:
     nanoseconds"""
     if not nanoseconds:
         return None
-    seconds, nanos = divmod(nanoseconds, 1e9)
+    seconds, nanos = divmod(nanoseconds, 10 ** 9)
```
aabmass, line 185: Can you create a constant like `NANOS_IN_SECONDS` that we have in the metrics exporter. We need to find a way to reuse code between the two exporters :/

aabmass: I'm assuming this gets converted into a proto `Timestamp` object eventually. Alternatively, you might be able to use `Timestamp.FromNanoseconds()` in place of this whole functions, e.g.
https://github.com/open-telemetry/opentelemetry-python/blob/5ff9600a62f9c82fc120b03e5729a53341465626/ext/opentelemetry-ext-opencensusexporter/src/opentelemetry/ext/opencensusexporter/util.py#L46

opentelemetry-exporter-cloud-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
```diff
@@ -182,7 +182,7 @@ def _get_time_from_ns(nanoseconds: int) -> Dict:
...
     return {"seconds": int(seconds), "nanos": int(nanos)}
```
aabmass, line 186: If `nanoseconds` is truly an int, `divmod()` will already give two ints

opentelemetry-exporter-cloud-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
```diff
@@ -182,8 +183,10 @@ def _get_time_from_ns(nanoseconds: int) -> Dict:
     nanoseconds"""
     if not nanoseconds:
         return None
-    seconds, nanos = divmod(nanoseconds, 1e9)
-    return {"seconds": int(seconds), "nanos": int(nanos)}
+    ts = Timestamp()
+    # pylint: disable=no-member
+    ts.FromNanoseconds(nanoseconds)
+    return ts
```
aabmass, line 189: Return type on L181 is out of date

PR #47, opened 2020-07-29
add support for list attribute values

AndrewAXue:
OT attribute values support having a sequence of primitives (list of ints, tuple of floats, etc). Cloud Trace does NOT support having a list of things as an attribute value, so we would ignore these values. This PR makes it so, instead of ignoring these, we convert them to a string and send it off as usual.

### DIFF
--- a/opentelemetry-exporter-cloud-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
+++ b/opentelemetry-exporter-cloud-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
@@ -40,6 +40,7 @@ API
 ---
 """
 
+import collections
 import logging
 from typing import Any, Dict, List, Optional, Sequence, Tuple
 
@@ -367,6 +368,7 @@ def _extract_attributes(
 
 
 def _format_attribute_value(value: types.AttributeValue) -> AttributeValue:
+
     if isinstance(value, bool):
         value_type = "bool_value"
     elif isinstance(value, int):
@@ -377,10 +379,15 @@ def _format_attribute_value(value: types.AttributeValue) -> AttributeValue:
     elif isinstance(value, float):
         value_type = "string_value"
         value = _get_truncatable_str_object("{:0.4f}".format(value), 256)
+    elif isinstance(value, collections.Sequence):
+        value_type = "string_value"
+        value = _get_truncatable_str_object(
+            "[" + ",".join([str(x) for x in value]) + "]", 256
+        )
     else:
         logger.warning(
             "ignoring attribute value %s of type %s. Values type must be one "
-            "of bool, int, string or float",
+            "of bool, int, string or float, or a sequence of these",
             value,
             type(value),
         )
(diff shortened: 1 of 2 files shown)


opentelemetry-exporter-cloud-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
```diff
@@ -377,10 +379,15 @@ def _format_attribute_value(value: types.AttributeValue) -> AttributeValue:
     elif isinstance(value, float):
         value_type = "string_value"
         value = _get_truncatable_str_object("{:0.4f}".format(value), 256)
+    elif isinstance(value, collections.Sequence):
```
aabmass, line 382: Did OT say specifically that only `Sequence` is allowed, or can it be any iterable (this will not pass for unordered iterables like set and dict)?

opentelemetry-exporter-cloud-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
```diff
@@ -377,10 +379,15 @@ def _format_attribute_value(value: types.AttributeValue) -> AttributeValue:
...
+        value_type = "string_value"
+        value = _get_truncatable_str_object(
+            "[" + ",".join([str(x) for x in value]) + "]", 256
+        )
```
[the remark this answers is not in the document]
AndrewAXue: Oh good point, I guess its obvious without the surrounding brackets anyways.

opentelemetry-exporter-cloud-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
AndrewAXue: OT SPec
```
AttributeValue = Union[
    str,
    bool,
    int,
    float,
    Sequence[str],
    Sequence[bool],
    Sequence[int],
    Sequence[float],
]
```
So the value shouldn't be a set or dict.

opentelemetry-exporter-cloud-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
```diff
@@ -377,10 +378,15 @@ def _format_attribute_value(value: types.AttributeValue) -> AttributeValue:
     elif isinstance(value, float):
         value_type = "string_value"
         value = _get_truncatable_str_object("{:0.4f}".format(value), 256)
+    elif isinstance(value, collections.Sequence):
+        value_type = "string_value"
+        value = _get_truncatable_str_object(
+            ",".join([str(x) for x in value]), 256
```
aabmass, line 384: nit, can use a generator
```suggestion
            ",".join(str(x) for x in value), 256
```

AndrewAXue: Ohhh very nice!

PR #48, opened 2020-07-29
add troubleshooting section for GKE resources

AndrewAXue:
A user ran into some trouble setting up the GKE resource detector. This troubleshooting section contains some common problems and solutions.

### DIFF
--- a/docs/examples/tools/cloud_resource_detector/README.rst
+++ b/docs/examples/tools/cloud_resource_detector/README.rst
@@ -55,7 +55,7 @@ You need to manually pass in some information via the `Downward API <https://kub
 to enable GKE resource detection. Your kubernetes config file should look
 something like this (passing in NAMESPACE, CONTAINER_NAME, POD_NAME)
 
-.. code-block:: sh
+.. code-block:: yaml
 
     apiVersion: "apps/v1"
     kind: "Deployment"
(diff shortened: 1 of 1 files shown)


docs/examples/tools/cloud_resource_detector/README.rst
```diff
@@ -44,3 +44,50 @@ After running the metrics example:
...
+
+
+Troubleshooting
+--------------------------
+
+gke_container resources are not being detected or exported:
+###########################################################
+You need to manually pass in some information via the `Downward API <https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/>`_
+to enable GKE resource detection. Your kubernetes config file should look
+something like this (passing in NAMESPACE, CONTAINER_NAME, POD_NAME)
```
aabmass, line 56:
```suggestion
something like this (passing in ``NAMESPACE``, ``CONTAINER_NAME``, ``POD_NAME``)
```

docs/examples/tools/cloud_resource_detector/README.rst
```diff
@@ -44,3 +44,50 @@ After running the metrics example:
...
+
+.. code-block:: sh
```
aabmass, line 58: should be yaml or yml? Not sure which one
```suggestion
.. code-block:: yml
```

AndrewAXue: After trying both `yml` and `yaml`, it seems they are the same, except `yaml` has colors. Going with the colorful one :)

PR #57, opened 2020-08-07
Exemplar support for Histogram aggregator

cnnradams:
I have this PR in draft state since exemplars are not merged in opentelemetry-python itself yet (the tests fail for the same reason), but exemplars can succesfully be uploaded to Cloud Monitoring with this code.

There is one TODO that needs to be addressed for uploading dropped labels alongside the trace/span context. I won't have time to add this support.


cnnradams:
Closing for now since there's no progress on the corresponding OT-python PR

Issue #19, opened 2020-07-13
Remove pinned "opentelemetry_python_commit" and test against releases


aabmass:
See #17 for discussion.

#17 has tox install opentelemetry-api/sdk packages to run tests against from a pinned commit. This keeps us up developing closely with opentelemetry-python repo before GA. After GA where there are fewer breaking changes, we should switch to actually testing against released versions of those packages.

PR #65, opened 2020-08-14
Adding mypy

### LINKED ISSUE #64: Set up mypy
Run mypy in dev/CI. Would be nice to have a config that works cleanly with vscode.

aabmass:
- mypy configs for each module
- tox setup to run mypi in CI
- output `junit.xml` files which CircleCI can use to display pretty failures
- type fixes/ignores where needed

Closes #64

Still need to run mypy on test code, will leave it for another PR.

### DIFF
--- a/dev-constraints.txt
+++ b/dev-constraints.txt
@@ -10,5 +10,5 @@ Sphinx==3.1.2
 # development before GA. After GA, we will build against specific releases.
 # Bump the commit frequently during development whenever you are missing
 # changes from upstream.
--e git+https://github.com/open-telemetry/opentelemetry-python.git@3cae0775ba12a2f7b4214b8b8c062c5e81002a19#egg=opentelemetry-api&subdirectory=opentelemetry-api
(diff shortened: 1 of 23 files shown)


dev-constraints.txt
```diff
@@ -10,5 +10,5 @@ Sphinx==3.1.2
 # development before GA. After GA, we will build against specific releases.
 # Bump the commit frequently during development whenever you are missing
 # changes from upstream.
-opentelemetry-api==0.12b0
-opentelemetry-sdk==0.12b0
+-e git+https://github.com/open-telemetry/opentelemetry-python.git@3cae0775ba12a2f7b4214b8b8c062c5e81002a19#egg=opentelemetry-api&subdirectory=opentelemetry-api
```
zoercai, line 13: OOC what is `@3cae0775ba12a2f7b4214b8b8c062c5e81002a19#egg=opentelemetry-api&subdirectory=opentelemetry-api`?

zoercai (approved):

dev-constraints.txt
aabmass: It installs opentelemetry-api/sdk packages from the git repo at that commit (`3cae0775ba12a2f7b4214b8b8c062c5e81002a19`) so we can track more closely than the latest release.

Planning to remove it after GA when there are fewer breaking changes: #19

PR #74, opened 2020-09-17
release pr/0.13b0

aabmass:
Please review the two commits individually (will be merged with "Rebase and Merge"). The first one will become the tagged release commit and the second is the master branch going forward:

- Release 0.13b0 (Part 1/2) release commit
- Release 0.13b0 (Part 2/2) bump version to 0.14.dev0

See release instructions: https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/blob/master/docs/releasing.md

### DIFF
--- a/dev-constraints.txt
+++ b/dev-constraints.txt
@@ -10,5 +10,5 @@ Sphinx==3.1.2
 # development before GA. After GA, we will build against specific releases.
 # Bump the commit frequently during development whenever you are missing
 # changes from upstream.
--e git+https://github.com/open-telemetry/opentelemetry-python.git@1abdc02a92cd996f3924e1653896bb72123ac84e#egg=opentelemetry-api&subdirectory=opentelemetry-api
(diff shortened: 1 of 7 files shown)


nilebox:
> The first one will become the tagged release commit and the second is the master branch going forward:

I don't think this will actually work, unless you merge with "fast forward".
See https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/merging-a-pull-request
> Pull requests are merged using the --no-ff option, except for pull requests with squashed or rebased commits, which are merged using the fast-forward option.

So the only case where "fast forward" is used is when you squash commits into one, which won't work for you either.
Seems like you need to send 2 separate PRs.

nilebox:
Looks like it does work via "Rebase and merge" somehow based on https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/blob/master/docs/releasing.md#open-and-merge-a-pr

LGTM then.

nilebox (approved):

zoercai (approved):

PR #83, opened 2020-10-23
fixed more 0.14b0 breakages and updated tests to catch them

aabmass:
There were some more breakages from v0.14b0 that needed fixing before release. Most of this PR is updating tests to actually catch the breakages in the future.
- Updated unit tests to use autospeccing
- Update the integration tests against GCM Mock api to catch the errors too
  * turns out the cloud monitoring integration tests were passing even though they shouldn't, because we swallow exceptions when `CreateMetricDescriptor()` fails
  * updated tests to fail if there is an error logged from `CreateMetricDescriptor()` failure
  * updated tests/exporter to actually pass
- Updated changelogs

### DIFF
--- a/opentelemetry-exporter-google-cloud/tests/test_cloud_monitoring_exporter_integration.py
+++ b/opentelemetry-exporter-google-cloud/tests/test_cloud_monitoring_exporter_integration.py
@@ -13,22 +13,24 @@
 # limitations under the License.
 
 
+import logging
+from unittest.mock import MagicMock, patch
+
 import grpc
 from google.cloud.monitoring_v3 import MetricServiceClient
 from google.cloud.monitoring_v3.gapic.transports import (
     metric_service_grpc_transport,
 )
 from opentelemetry.exporter.cloud_monitoring import (
-    NANOS_PER_SECOND,
-    WRITE_INTERVAL,
     CloudMonitoringMetricsExporter,
 )
 from opentelemetry.sdk import metrics
-from opentelemetry.sdk.metrics.export import MetricRecord, MetricsExportResult
-from opentelemetry.sdk.metrics.export.aggregate import SumAggregator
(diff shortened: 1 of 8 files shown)


opentelemetry-exporter-google-cloud/tests/test_cloud_monitoring_exporter_integration.py
```diff
@@ -13,56 +13,70 @@
...
 
-        meter = metrics.MeterProvider().get_meter(__name__)
+        meter_provider = metrics.MeterProvider(
+            resource=Resource.create(
+                {
+                    "cloud.account.id": "some_account_id",
+                    "cloud.provider": "gcp",
+                    "cloud.zone": "us-east1-b",
+                    "host.id": 654321,
+                    "gcp.resource_type": "gce_instance",
```
nilebox, line 54: Have we decided to keep this one? It's incompatible with semantic conventions, but might be okay as an additional attribute.

Also, IIRC you shared a link where AWS has introduced attributes / resources specific to AWS to the spec? Shall we try doing the same if we want to support GCP-specific resource attributes?

nilebox (approved):

opentelemetry-exporter-google-cloud/tests/test_cloud_monitoring_exporter_integration.py
aabmass: > Have we decided to keep this one? It's incompatible with semantic conventions, but might be okay as an additional attribute.

I think it's better to use the semantic conventions if we aren't missing any functionality https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/issues/85. Do you have any preference on adding our own GCP resource attributes instead? (Here's the AWS PR for reference https://github.com/open-telemetry/opentelemetry-specification/pull/1099))

Issue #97, opened 2021-01-28
0.17b0? / PyPI release


twoodwark:
Hi there, I see https://pypi.org/project/opentelemetry-exporter-google-cloud/ is still on 0.16b1.  Is there any chance of a new release? 

I'd find it really useful to have 0.17b0 there.  

I'm intending to add a conda-forge feedstock for this package, but other packages like https://anaconda.org/conda-forge/opentelemetry-api are already at opentelemetry version 0.17b0,  skipping 0.16b1.

aabmass:
Thanks for reporting, cutting a new release now

PR #101, opened 2021-02-03
upgrade OTel python to 0.17b0, small fixes

aabmass:
- Upgrade to latest python SDK
- Add fields
- Typing improvements

Header checker is failing because the modified files are copyrighted to OpenTelemetry Authors.

#97

### DIFF
--- a/.github/header-checker-lint.yml
+++ b/.github/header-checker-lint.yml
@@ -1,5 +1,8 @@
 allowedLicenses:
   - Apache-2.0
+allowedCopyrightHolders:
+  - Google
+  - OpenTelemetry Authors
 sourceFileExtensions:
   - py
   - sh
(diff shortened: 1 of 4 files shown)


.github/header-checker-lint.yml
```diff
@@ -1,5 +1,8 @@
 allowedLicenses:
   - Apache-2.0
+allowedCopyrightHolders:
+  - Google
+  - OpenTelemetry Authors
```
aabmass, line 5: Are we in favor of adding this? I want the checker to pass so it remains a good signal, even though its not ideal to add it

opentelemetry-exporter-google-cloud/src/opentelemetry/exporter/cloud_trace/__init__.py
```diff
@@ -411,7 +413,7 @@ def _extract_attributes(
         )
     return ProtoSpan.Attributes(
         attribute_map=attributes_dict,
-        dropped_attributes_count=attributes_dict.dropped
+        dropped_attributes_count=attributes_dict.dropped  # type: ignore[attr-defined]
```
punya, line 416: As a follow up, should `dropped` be added to the `BoundedDict` type definition so that this pragma can be removed?

punya (approved):
LGTM

opentelemetry-exporter-google-cloud/src/opentelemetry/exporter/cloud_trace/__init__.py
aabmass: Created an issue https://github.com/open-telemetry/opentelemetry-python/issues/1573

PR #111, opened 2021-03-15
Fix opentelemetry-tools-google-cloud homepage URL

ymotongpoo:
Current URL mentioned in PyPI website is referring to non-existing URL and returns 404. This corrects the URL to the corresponding page of the directory in the GitHub repository.


aabmass (approved):
Thanks!

PR #112, opened 2021-03-18
Handle mixed case cloud propagator headers

### LINKED ISSUE #109: Cloud Trace Propagator Doesn't Work with gRPC
The Cloud Trace Propagator uses the `X-Cloud-Trace-Context` header key, which has upper case letters. When using it with the gRPC instrumentation, it sets this as gRPC metadata which will throw a validation error and abort the request because of the upper case letters.

Please make the key all lower case, I tested it on my own services on Cloud Run using a Flask app (http from Google Cloud's Load Balancer which sets the x-cloud-trace-context header initially), and is passed on to my backend service using gRPC and it works with the metadata key set to the same, and appears that the AppServer component on Cloud Run in-between the two services also picks up the trace ID with the lowercase metadata header, and everything links up correctly as expected so I can now see my full end-to-end request on Google Trace correctly.

aabmass:
Fixes #109
- Extract:
  - First look for all lowercase `x-cloud-trace-context`
  - If not found, search all other headers for mixed case match
- Inject: always write `x-cloud-trace-context` (lowercase) on outgoing propagated requests

In addition to this change, I am investigating if OpenTelemetry can handle the extract conversion internally so that each instrumentation/propagator doesn't have to handle this sort of conversion, when most carriers are mixed case.


ymotongpoo (approved):
For gRPC compatibility, LGTM

PR #124, opened 2021-04-07
split resource tools into opentelemetry-propagator-gcp and opentelemetry-resourcedetector-gcp

aabmass:
Split `opentelemetry-tools-google-cloud` package into separate
- `opentelemetry-propagator-gcp`
- `opentelemetry-resourcedetector-gcp`


.circleci/config.yml
```diff
@@ -58,18 +58,18 @@ jobs:
     steps:
       - setup
       - run_with_pip_cache:
-          cache_key_base: lint-python3.8
+          cache_key_base: lint-python3.9
           steps:
-            - run: tox -f lint-ci
+            - run: tox -s false -f lint-ci
```
punya, line 63: Nit: When writing scripts or defining CI steps, consider using long-form command line flags for readability (I had to look up that this meant `--skip-missing-interpreters`).

punya (approved):

.circleci/config.yml
punya: This is non-blocking feedback. There are lots of single-character flags used already, so you don't have to fix them all for this PR. It's okay to fix just this one, or plan to fix them all later in one fell swoop.

aabmass: 👍 good point

opentelemetry-exporter-gcp-trace/src/opentelemetry/exporter/cloud_trace/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.18.dev0"
+__version__ = "0.19.dev0"
```
aabmass, line 15: Missed this in a previous PR somehow

punya: Do we have tooling to automate editing this and/or enforce consistency across the different packages?

PR #128, opened 2021-04-12
add flask example and region tags

aabmass:
- add region tags to advanced_trace.py
- add flask e2e example with new region tags


aabmass:
Built docs for the new flask example from the CircleCI run: https://3200-244484614-gh.circle-artifacts.com/0/docs-html/examples/flask_e2e/README.html

aabmass:
> Here is the summary of possible violations 😱
> 
> There are 9 possible violations for not having product prefix.
> **The end of the violation section. All the stuff below is FYI purposes only.**

This is flaky and OK to ignore right now. I verified there is a product prefix

punya (approved):

PR #132, opened 2021-04-21
update release.py to work with different package versions

aabmass:
The `release.py` script currently keeps all packages at the same version. This PR changes it to accept a JSON config for which versions to change to what, since metrics and resource detection are now at different versions. See updated releasing.md for instructions.


aabmass:
Going to take a simpler approach that they all have same version numbers but can append different suffix

PR #134, opened 2021-04-22
Update releasing script to allow specifying suffixes for different packages

aabmass:
- Remove pinning stuff, as its no longer needed
- Can now specify alternate suffixes for certain packages. This is in line with what our ops-java repo does, where metrics and resource detector are marked alpha, but otherwise the same version. See updated `releasing.md` for explanation
- Verify everything is PEP440/pypi compliant now

Release PR generated from these changes here: https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/pull/135

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -50,21 +50,13 @@ the repository to `0.12.dev0`:
     --new_dev_version 0.12.dev0
 ```
 
-You can also specify alternative suffixes to add for certain packages. For
-example, to release all packages as `1.0.0rc0` except
-`opentelemetry-exporter-gcp-monitoring` and
-`opentelemetry-resourcedetector-gcp` as `1.0.0a0`:
+You can also specify alternative suffixes to add for certain packages by
+updated the `ALTERNATE_SUFFIXES` map in `release.py`. This lets you mark some
+packages as rc/alpha/beta/etc. Check the code comments for details.
 
-```bash
-./release.py \
-    --release_version 1.0.0rc0 \
-    --new_dev_version 1.1.0.dev0 \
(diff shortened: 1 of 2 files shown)


punya:
How about having a config file that specifies which packages get the alternate suffix and which don't? Otherwise, someone has to construct that long command line each time.

aabmass:
> How about having a config file that specifies which packages get the alternate suffix and which don't? Otherwise, someone has to construct that long command line each time.

Let me give it a shot to see how it looks

docs/releasing.md
```diff
@@ -95,23 +109,17 @@ buggy) small script for the example PR's tag:
...
 ```bash
 # pull in the new release tag
 git fetch origin
 
-# create branch and move stable
-git branch release/0.11b0 v0.11b0
+# move stable
 git tag -d stable
 git push origin :refs/tags/stable
-git tag stable v0.11b0
```
aabmass, line 110: add back

punya:
It could be as simple as a map hardcoded in `release.py`.

aabmass:
Thanks @punya it looks much better now

docs/releasing.md
```diff
@@ -41,17 +41,23 @@ branch. The script does the following:
...
 ```bash
 ./release.py \
     --release_version 0.11b1 \
-    --new_dev_version 0.12.dev0 \
-    --ot_version "==0.11b0"
+    --new_dev_version 0.12.dev0
 ```
 
+You can also specify alternative suffixes to add for certain packages by
+updated the `ALTERNATE_SUFFIXES` map in `release.py`. This lets you mark some
```
punya, line 54: *updating

punya (approved):

PR #136, opened 2021-04-29
test server for python e2e tests

### LINKED ISSUE #41: Add more integration tests for cloud trace
Some things we'd like to test are:
- Tests for the various features that the Exporter supports. (events, links, status codes, etc.)
- Test that contains an invalid span and assert failure
- Export multiple spans

aabmass:
Fixes #41

- add instrumented test server and docker file for it
- add cloudbuild.yaml and created a trigger to actually run the tests

See the passing checks


cloudbuild.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+steps:
+  - name: "docker"
+    args:
+      - build
+      - --tag=gcr.io/$PROJECT_ID/opentelemetry-operations-python-e2e-test-server
+      - --file=e2e-test-server/Dockerfile
+      - .
+
+  - name: gcr.io/opentelemetry-ops-e2e/opentelemetry-operations-e2e-testing:0.5.5
```
jsuereth, line 23: Nit: Do we want the name that shows up to be this, or just "Push image to GCR"

e2e-test-server/server.py
```diff
@@ -0,0 +1,72 @@
...
+@contextlib.contextmanager
+def common_setup() -> Iterator[tuple[str, Tracer]]:
+    """\
+    Context manager with common setup for test endpoints
+
+    It extracts the test-id header, creates a tracer, and finally flushes
+    spans created during the test
+    """
+
+    if TEST_ID not in request.headers:
```
jsuereth, line 41: Does this imply we don't set up the tracer provider until the first URL comes in OR that we set one up for EVERY request to the server?

cloudbuild.yaml
aabmass: So `name` is (confusingly) the container image to run: https://cloud.google.com/build/docs/build-config#name. This step runs the test server, I'm not pushing the built image to GCR at all actually.

I think [`id`](https://cloud.google.com/build/docs/build-config#id) changes what shows in the UI, but it also used to configure dependencies. Lets me check

e2e-test-server/server.py
aabmass: We set one up for every request to the server. So test has its own provider/processor/exporter for isolation

cloudbuild.yaml
aabmass: Updated. I used kebab strings for ID so its clear they are not primarily human readable. Now the build UI looks like this

Screen Shot 2021-05-03 at 2 11 22 PM

jsuereth: Ah, that looks good.

e2e-test-server/server.py
jsuereth: Interesting.  Is that a requirement for other languages?

E.g. would it be ok to just "flush" after serving the trace?

aabmass: It is in the design, but flush would work too. I figured each test scenario might as well have a completely separate SDK setup.

Do you have a preference for flush?

jsuereth: No, just worried about other languages.  What's here is good for now.

jsuereth (approved):

e2e-test-server/server.py
aabmass: Talked offline – each BatchSpanProcessor creates a new thread. But, calling `tracer_provider.shutdown()` will cause the thread to end and wait block until `Thread.join()` finishes.

Updated the spec that this is not a requirement but a suggestion.

PR #145, opened 2021-05-19
Add option for additional resource attributes with regex to trace exporter

dhpollack:
The [OpenTelemetry Specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#specifying-resource-information-via-an-environment-variable) allows one to specify additional attributes with the special environmental variable `OTEL_RESOURCE_ATTRIBUTES`.  This includes those attributes into the span exported to GCP.

The goal of this PR is to easy add "global" attributes to all spans in a given service so that they can be filtered easily in GCP.  But also it is not intuitive that you can add these attributes to spans, but they don't get exported with the `CloudTraceSpanExporter`.

Signed-off-by: David Pollack <david@da3.net>


aabmass:
Can you remove the `OTEL_RESOURCE_ATTRIBUTES` parts since that is already covered by OTel python? We are thinking the other part of this PR would be great to add, with some configuration. Can you update the `CloudTraceSpanExporter` constructor to accept a regex for the resource attributes to pass through. If the option isn't given, it shouldn't change the existing behavior

opentelemetry-exporter-gcp-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
```diff
@@ -357,24 +364,32 @@ def _strip_characters(ot_version):
...
-            resource_attributes[ot_resource_key]
-        )
-        for ot_resource_key, gcp_resource_key in OT_RESOURCE_ATTRIBUTE_TO_GCP[
-            resource_type
-        ].items()
-    }
+        return extracted_attributes
+    extracted_attributes.update(
+        {
+            "g.co/r/{}/{}".format(resource_type, gcp_resource_key,): str(
```
aabmass, line 384: Only certain resource attributes are recognized by Google with this `g.co/r/` format. The ones that pass the regex can just be added straight in without changing the keys.

opentelemetry-exporter-gcp-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
```diff
@@ -88,16 +89,22 @@ class CloudTraceSpanExporter(SpanExporter):
...
+        self, project_id=None, client=None, use_otel_env_resource=False,
     ):
         self.client = client or TraceServiceClient()
         if not project_id:
             _, self.project_id = google.auth.default()
         else:
             self.project_id = project_id
+        self.otel_env_resource = (
+            OTELResourceDetector().detect() if use_otel_env_resource else None
+        )
```
aabmass, line 107: This isn't needed, this is done automatically by OTel. The resource this exporter sees below will already have the fields from the environrment variable.

I was thinking we could just add a regex here, where resource keys that match the regex will be passed on as span attributes.

aabmass (changes requested):

opentelemetry-exporter-gcp-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
dhpollack: But we don't know what the resource keys are because they are based on the environmental variable.  So you want only a few pre-set list of resource attribute to be passed on to the span attribute?  Doesn't that kind of defeat the purpose of allowing people to use the `OTEL_RESOURCE_ATTRIBUTES` environmental variable?

opentelemetry-exporter-gcp-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
dhpollack: This was extracted [directly from `main`](https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/blob/master/opentelemetry-exporter-gcp-trace/src/opentelemetry/exporter/cloud_trace/__init__.py#L370).  I can change this, but what format at you looking for exactly.

opentelemetry-exporter-gcp-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
dhpollack: I could just extract the keys here, but I thought it made a little more sense for all the inputs to `_extract_resources` to be resources instead of a hybrid between `Resource` and `Attribute` / `Dict[str, str]`.  The other idea is to add it to `resources_and_attrs` in line 166 directly with `**self.otel_env_resource`, but then testing would require a bit of refactoring.

dhpollack: In my use case, I gave some examples to simplify what I actually wanted.  The truth is we use a non-standard name called `app_name` and I would probably use that instead of `service.name`.  But I didn't want this PR to be able my specific case, but more about generally adding extra attributes to the spans.  The idea being that one could add arbitrary attributes (even ridiculously named non-standard attributes like `app_name`) that would show up in the GCP interface as labels which can be used for filtering by using the environmental variable `OTEL_RESOURCE_ATTRIBUTES`.

Apologies if I wasn't clear on this earlier.

aabmass: No worries. For the general use case, I think we can address it again later if someone can't solve it with a more restrictive regex.

Would either a filter function that returns true/false if the resource key should be included, or a regex pattern for resource keys to include solve your specific issue?

opentelemetry-exporter-gcp-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
aabmass: The `g.co/r/` attributes should only correspond to monitored resource labels (e.g. [gce_instance](https://cloud.google.com/monitoring/api/resources#tag_gce_instance)). The additional resource labels we are putting can just be copied over from the OTel resource in the span.

opentelemetry-exporter-gcp-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
dhpollack: Ok, I get it.  Yea, that would work for me.  I'll implement the regex.

opentelemetry-exporter-gcp-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
dhpollack: @aabmass let me know if this is what you want now.  For `gce_instance` I kept it the same as before but for `gke_container` I only use the `ot_resource_key` (for example `k8s.cluster.name`).

dhpollack:
@aabmass I think this is ready now

opentelemetry-exporter-gcp-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
```diff
@@ -88,16 +89,21 @@ class CloudTraceSpanExporter(SpanExporter):
         project_id: ID of the cloud project that will receive the traces.
         client: Cloud Trace client. If not given, will be taken from gcloud
             default credentials
+        resource_regex: A regex string to match and extract resource
+            attributes from spans (default: None).
```
aabmass, line 93:
```suggestion
        resource_regex: Resource attributes with keys matching this regex will be
          added to exported spans as labels (default: None).
```

opentelemetry-exporter-gcp-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
```diff
@@ -357,24 +363,38 @@ def _strip_characters(ot_version):
...
-        for ot_resource_key, gcp_resource_key in OT_RESOURCE_ATTRIBUTE_TO_GCP[
-            resource_type
-        ].items()
-    }
+        return extracted_attributes
+    extracted_attributes.update(
+        {
+            "g.co/r/{}/{}".format(resource_type, gcp_resource_key,)
+            if resource_type == "gce_instance"
+            else ot_resource_key: str(resource_attributes[ot_resource_key])
```
aabmass, line 391: I don't think this if/else is needed, we still special case both keys of `OT_RESOURCE_ATTRIBUTE_TO_GCP`.

opentelemetry-exporter-gcp-trace/tests/test_cloud_trace_exporter.py
```diff
@@ -621,6 +622,54 @@ def test_too_many_link_attributes(self):
     def test_extract_empty_resources(self):
         self.assertEqual(_extract_resources(Resource.get_empty()), {})
 
+    def test_extract_resource_attributes_with_regex(self):
+        resource_regex = re.compile(r"service\..*")
```
aabmass, line 626: I notice we are using `regex.match()` above, in which case the `.*` isn't required here. What do you think is more useful/intuitive, `fullmatch()` or `match()`?

aabmass (commented):
Nice, looks good! Just add back the gke_cluster case and we should be good

opentelemetry-exporter-gcp-trace/tests/test_cloud_trace_exporter.py
dhpollack: I personally don't use python regexes that often, so I wouldn't read too much into my intention based on this regex (i.e. this is probably an inefficient / "bad" regex because I don't write them that often).  As a general rule, I would use which ever one seems more prevalent.  But I would certainly defer to you as to what you prefer on this one.

opentelemetry-exporter-gcp-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
dhpollack: ok, I misread you previous comment.  changed this back.

opentelemetry-exporter-gcp-trace/tests/test_cloud_trace_exporter.py
aabmass: I think the same sort of thing could be accomplished with all of them. Lets leave it as match for now 👍

aabmass (approved):
Thanks for working through my comments, LGTM!

aabmass:
@dhpollack could you rename this PR to something more descriptive of what we landed on, and add an entry to the changelog file? I'll merge after that 😃

dhpollack:
> @dhpollack could you rename this PR to something more descriptive of what we landed on, and add an entry to the changelog file? I'll merge after that smiley

@aabmass done!  Thanks for all the help.

PR #151, opened 2021-06-01
Upgrade dev dependencies and reformat the code

aabmass:
Also, pin them

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,16 +1,6 @@
 [settings]
-include_trailing_comma=True
-force_grid_wrap=0
-use_parentheses=True
+profile=black
 line_length=79
-
-; 3 stands for Vertical Hanging Indent, e.g.
-; from third_party import (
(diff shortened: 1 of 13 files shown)


.isort.cfg
```diff
@@ -1,16 +1,6 @@
 [settings]
-include_trailing_comma=True
-force_grid_wrap=0
-use_parentheses=True
+profile=black
```
aabmass, line 2: New isort version has a profile for compat with the black autoformatter, so switching to that.

tox.ini
```diff
@@ -101,7 +101,7 @@ changedir = {env:PACKAGE_NAME}
 commands =
   ; To run fixers (black, isort) `tox -f fix`.
   fix: black .
-  fix: isort --recursive .
```
aabmass, line 104: this flag was deprecated

dashpole (approved):

Issue #172, opened 2021-12-28
gcp cloud trace and OTEL ResourceAttributes not working


raphaelauv:
I can't find the service.name in the filters of the gcp cloud trace UI

I gave a full example there https://github.com/GoogleCloudPlatform/python-docs-samples/issues/7300

How can I filter by OTEL service.name in cloud trace ?

aabmass:
@raphaelauv I think there are two separate issues here.

The `service.name` attribute isn't sent by default, you will need to allowlist it with this regex option: https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/blob/ee23dfa616300d2762c3c974f3e2881f0a206d0b/opentelemetry-exporter-gcp-trace/src/opentelemetry/exporter/cloud_trace/__init__.py#L101-L102

So something like `r"service\.name"` or a more permissive regex will cause it to be added to a span label. Note this is currently unreleased, but expect a release soon.

The second issue is filtering in the UI. As you found out, the "Service" option in the filter bar is not what you want. Once the label is being sent, you can use the filter syntax you probably already saw in the docs [here](https://cloud.google.com/trace/docs/trace-filters#filter_syntax).

aabmass:
Just released version 1.1.0 with the `resource_regex` parameter. Closing for now but let me know if you have any issues.

raphaelauv:
in GO it's automatically manage : https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/119/files

could we add this feature to python ? @aabmass

PR #182, opened 2022-04-14
Support proper Service mapping

AndreasBergmeier6176:
To keep up with Go implementation and be able to see Service in Trace UI.


aabmass:
@dashpole is this something we should be implementing in all the trace exporters?

AndreasBergmeier6176:
Is there a TODO still open for this to get merged?

raphaelauv:
close : https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/issues/172

raphaelauv:
any news?

dashpole:
This isn't needed anymore.  The Trace UI now will use service.name for the Service without this mapping.

PR #186, opened 2022-05-06
Update no container name behaviour

munagekar:
- If the env `CONTAINER_NAME` is not found, it is not set, instead of silently disabling the resource detector all together.
- This makes the behavior consistent with https://github.com/GoogleCloudPlatform/opentelemetry-operations-go.


munagekar:
@aabmass @dashpole PTAL.

dashpole:
This lgtm, but i'll let @aabmass approve

munagekar:
@aabmass Could you take a look ?

munagekar:
@dashpole @aabmass Sorry for the ping again. Just pinging in case my previous message went unnoticed.

aabmass (approved):

PR #187, opened 2022-05-16
Use experimental metrics api

ksmets:
As I wanted to enable both tracing and monitoring in one project, I modified the code to use the 
experimental metrics api. 

Keep in mind that I made the following assumptions:
- keep the start and end times of the exported metrics as is
- bounds and values of histogram are of equal length

Moreover I added locking when accessing the metric descriptors (otherwise `test_cloud_monitoring_exporter_integration.py` would fail randomly).
I introduced `metrictestutil.py` with convenience functions to create metrics (based on https://github.com/open-telemetry/opentelemetry-python/blob/main/tests/opentelemetry-test-utils/src/opentelemetry/test/metrictestutil.py)

### DIFF
--- a/docs/examples/cloud_monitoring/basic_metrics.py
+++ b/docs/examples/cloud_monitoring/basic_metrics.py
@@ -15,27 +15,30 @@
 
 import time
 
-from opentelemetry import metrics
+from opentelemetry._metrics import get_meter_provider, set_meter_provider
 from opentelemetry.exporter.cloud_monitoring import (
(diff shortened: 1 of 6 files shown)


aabmass:
Thanks for taking a stab at this. I've opened a separate PR https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/pull/204 and decided to gut some of the code and tests as so much of it is no longer applicable.

Issue #85, opened 2020-10-26
Remove "gcp.resource_type" resource attribute


aabmass:
Instead, use OTel resource semantic conventions. See Go reference impl https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/402639db27fdf8e5143cfd8a871d40cbe10d4bd2/exporter/metric/metric.go#L365

aabmass:
This was fixed in https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/pull/214

PR #194, opened 2022-08-23
Resource Detector for Cloud Run/Cloud Functions

### LINKED ISSUE #193: Resource detector should detect more services
The [Go resource detector](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/detectors/gcp/detector.go#L38) detects a lot of GCP services, like Cloud Run and Cloud Functions, while the python one only detects GCE and GKE.

RangelReale:
Based on the [Go](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/detectors/gcp/detector.go#L38) code.

Fixes #193

### DIFF
--- a/opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/__init__.py
+++ b/opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/__init__.py
@@ -112,11 +112,11 @@ def get_cloudrun_resources():
 
     faas_name = os.getenv("K_SERVICE")
     if faas_name is not None:
-        common_attributes["faas.name"] = faas_name
+        common_attributes["faas.name"] = str(faas_name)
 
     faas_version = os.getenv("K_REVISION")
     if faas_version is not None:
-        common_attributes["faas.version"] = faas_version
+        common_attributes["faas.version"] = str(faas_version)
 
     common_attributes.update(
         {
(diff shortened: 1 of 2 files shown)


aabmass (approved):
LGTM, @dashpole can you approve if the resource detection logic looks good?

opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/__init__.py
```diff
@@ -99,12 +99,74 @@ def get_gke_resources():
...
+    faas_version = os.getenv("K_REVISION")
+    if faas_version is not None:
+        common_attributes["faas.version"] = str(faas_version)
+
+    common_attributes.update(
+        {
+            "cloud.platform": "gcp_cloud_run",
+            "cloud.region": all_metadata["instance"]["region"].split("/")[-1],
+            "faas.id": all_metadata["instance"]["id"],
+            "gcp.resource_type": "cloud_run",
```
dashpole, line 126: I'm not familiar with `gcp.resource_type`.  It seems like duplicate information with cloud.platform.  Why is it included?

opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/__init__.py
```diff
@@ -99,12 +99,74 @@ def get_gke_resources():
...
+    faas_version = os.getenv("K_REVISION")
+    if faas_version is not None:
+        common_attributes["faas.version"] = str(faas_version)
+
+    common_attributes.update(
+        {
+            "cloud.platform": "gcp_cloud_functions",
+            "cloud.region": all_metadata["instance"]["region"].split("/")[-1],
+            "faas.id": all_metadata["instance"]["id"],
+            "gcp.resource_type": "cloud_functions",
```
dashpole, line 156: same as above.

opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/__init__.py
aabmass: It predates `cloud.platform` so it was our hack to make things work. We should get rid of it, see #85

dashpole:
Going to merge this, and then track removal of the extra attribute with #85

PR #197, opened 2022-09-06
fix auto-instrument loading tests for trace exporter.

muncus:
call to `_import_exporters()` is missing a positional arg, as reported by [this failing PR test](https://app.circleci.com/pipelines/github/GoogleCloudPlatform/opentelemetry-operations-python/731/workflows/8cc32f9e-88f5-4e1e-bc2f-66f8c52a8736/jobs/4816) also verified locally on a clean checkout.

### DIFF
--- a/cloudbuild-e2e-gce.yaml
+++ b/cloudbuild-e2e-gce.yaml
@@ -30,6 +30,7 @@ steps:
     args:
       - gce
       - --image=$_TEST_SERVER_IMAGE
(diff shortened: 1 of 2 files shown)


aabmass (approved):
/gcbrun

aabmass:
Looking at the e2e test log, the VM is taking longer to start than the timeout we expect. Let me tweak it to get the tests passing.

PR #198, opened 2022-09-15
Update gce build to have a higher health check timeout

aabmass:
See https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/pull/197#issuecomment-1248536459


aabmass:
Since this fix is circularly dependent on #197, they will either have to be squashed together or one force-merged. I think I will push to the other branch.

PR #203, opened 2022-11-03
Drop support for Python 3.6, add 3.10 and 3.11

aabmass:
This follows the supported versions in upstream OTel python.


aabmass (commented):
~Oops, need to update the setup.cfg files~ done

punya (approved):

punya:
You'll also need to remove the required check for tests-python3.6.

PR #205, opened 2022-11-04
Drop python3.11 temporarily

aabmass:
Until https://github.com/googleapis/proto-plus-python/issues/326 is fixed, client libraries don't properly support 3.11

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -120,4 +120,4 @@ workflows:
           name: tests-python3.<< matrix.python3_minor >>
           matrix:
             parameters:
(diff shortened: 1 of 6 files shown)


punya:
Should we temporarily mark that check as not required, instead of removing it altogether? (Note we'll have to mark it as not required even if we merge this PR.)

aabmass:
I'm hoping to release this code soon, probably before this issue is fixed. I don't want to claim 3.11 is supported if it wont' actually work

aabmass:
@punya if it looks OK, can you approve?

punya (approved):

PR #216, opened 2022-11-16
Set gRPC user-agent when calling google APIs

### LINKED ISSUE #63: Missing user agent when calling Monitoring/Trace apis
This is implemented in the Go exporter already, see [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/5fb47d1fa010a14731ca4fcee9bd5e57b25bd5b2/exporter/metric/option.go#L29) and [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/7454a7ef9c1361afe6b8b2d33e4b85a67c5db897/exporter/metric/metric.go#L124). Note this is separate from the `g.co/agent` span label set [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/blob/3f342f081b3f7ccdc52f7bdac716c8551a8033fe/opentelemetry-exporter-gcp-trace/src/opentelemetry/exporter/cloud_trace/__init__.py?q=user_agent#L489)

aabmass:
Fixes #63

Adds user-agent to gRPC calls. Note this will not work if the user overrides the client on their own.

Added an integration test which checks the actual `user-agent` metadata sent in gRPC requests


opentelemetry-exporter-gcp-monitoring/tests/test_user_agent.py
```diff
@@ -0,0 +1,39 @@
...
+        "mycounter", description="foo", unit="{myunit}"
+    )
+    counter.add(12)
+    meter_provider.force_flush()
+
+    for calls in gcmfake.get_calls().values():
+        for call in calls:
+            assert (
+                re.match(
+                    r"^opentelemetry-python \S+; google-cloud-metric-exporter \S+ grpc-python/\S+",
```
aabmass, line 35: @punya does this look good to you? I think the collector exporter just sets `opentelemetry-collector-contrib` but this seems a little more useful.

Lmk if you prefer a different format

dashpole: This is what we do for go: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/fbfa6aa3082c07d81ae7ad61e8ccdea9c85b63f6/exporter/metric/option.go#L29

`opentelemetry-go %s; google-cloud-metric-exporter %s`

aabmass:
I dropped a comment here https://github.com/googleapis/python-api-core/issues/341#issuecomment-1315943371. What are your thoughts here in terms of submitting this vs pushing on upstream libraries @punya?

punya:
I'd like us to start by proposing the upstream change that we'd ideally want. If we encounter resistance, we should fall back to this approach.

@jsuereth and @dashpole, curious to hear your thoughts too.

dashpole:
I'd prefer the reverse.  Make it work here, and open issues (here and upstream, if one doesn't already exist) to track making it cleaner/easier upstream.

aabmass:
@punya @dashpole can we go ahead with this PR?

punya (approved):
LGTM

PR #231, opened 2023-02-16
Add spec compliant GCE detection

aabmass:
Mostly just copied from the JS implementation. Note Google doesn't provide a metadata library we can use here directly like we have in Go and JS.

After I have implemented GKE detection, I will replace the existing (unstable) impl with the spec compliant one and update the trace exporter and e2e tests.

### DIFF
--- a/opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_gce.py
+++ b/opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_gce.py
@@ -52,7 +52,7 @@ def host_id() -> str:
 
 
 def host_name() -> str:
-    return get_metadata()["instance"]["hostname"]
+    return get_metadata()["instance"]["name"]
 
 
 @dataclass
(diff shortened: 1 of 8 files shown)


opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_gce.py
```diff
@@ -0,0 +1,75 @@
...
+def host_type() -> str:
+    return get_metadata()["instance"]["machineType"]
+
+
+def host_id() -> str:
+    return get_metadata()["instance"]["id"]
+
+
+def host_name() -> str:
+    return get_metadata()["instance"]["hostname"]
```
dashpole, line 55: I think we want instance/name instead of instance/hostname

aabmass: Ah nice catch. I'll have to fix in JS as well https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/786cec33b16a0b8c776daa00c6771320bc217b21/packages/opentelemetry-resource-util/src/detector/gce.ts#L66

dashpole (approved):

PR #233, opened 2023-02-17
Add spec compliant GKE detection

aabmass:


opentelemetry-resourcedetector-gcp/tests/test_gcp_resource_detector.py
```diff
@@ -95,20 +95,16 @@ def pop_environ_key(key):
         os.environ.pop(key)
 
 
-def clear_gke_env_vars():
```
aabmass, line 98: This file is the test for the existing resource detector which is being replaced.

It was leaking changes to `os.environ` into the other tests which I had to fix

opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_gke.py
```diff
@@ -0,0 +1,58 @@
...
+    _metadata,
+)
+
+# TODO: remove when Python 3.7 is dropped
+from typing_extensions import Literal
+
+KUBERNETES_SERVICE_HOST_ENV = "KUBERNETES_SERVICE_HOST"
+
+
+def on_gke() -> bool:
```
jsuereth, line 29: Should this also check for the presence of the metadata server?

opentelemetry-resourcedetector-gcp/tests/test_gke.py
```diff
@@ -0,0 +1,78 @@
...
+
+# Reset stuff before every test
+# pylint: disable=unused-argument
+@pytest.fixture(autouse=True)
+def autouse(fake_get_metadata, fake_environ):
+    pass
+
+
+def test_detects_on_gke() -> None:
+    os.environ["KUBERNETES_SERVICE_HOST"] = "fakehost"
```
jsuereth, line 30: Shouldn't we also verify we have access to the metadata server here?

opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_gke.py
dashpole: The metadata server is assumed to be available at this point in the detector.  See https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/pull/233/files#diff-690ac73bc4cd4aba0fbc7ff5f2233fb14ec3c86c407f07b6b39b22dc7718aa45R53-R54.

dashpole (commented):
lgtm.  Not approving just in-case it auto-merges

opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_gke.py
aabmass: I'm OK to sprinkle the check throughout the `on_x()` methods since it's cached anyway.

Fwiw, I'm not planning to expose these functions (remember we thought the GCP Logging client libs might use these but things changed).

jsuereth: Ah, I missed the not metadata check (and I had looked for it).

jsuereth (approved):

opentelemetry-resourcedetector-gcp/tests/test_gke.py
aabmass: I think the discussion above applies here

PR #234, opened 2023-02-21
Implement resource detection e2e scenario

aabmass:
It is successfully detecting on GKE, GCE. Will add GCR and other serverless in follow up PRs.

### DIFF
--- a/e2e-test-server/e2e_test_server/scenarios.py
+++ b/e2e-test-server/e2e_test_server/scenarios.py
@@ -14,12 +14,16 @@
 
 import contextlib
 import dataclasses
-from typing import Callable, Iterator, Mapping
+import os
+from typing import Any, Callable, Iterator, Mapping
 
 from google.rpc import code_pb2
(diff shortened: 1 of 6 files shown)


e2e-test-server/e2e_test_server/scenarios.py
```diff
@@ -124,6 +135,36 @@ def basic_propagator(request: Request) -> Response:
...
+
+    with _tracer_setup(
+        tracer_provider_config={
+            "resource": GoogleCloudResourceDetector(
+                raise_on_error=True
+            ).detect()
+        },
+        exporter_config={"resource_regex": r".*"},
+    ) as tracer:
+        propagator = CloudTraceFormatPropagator()
```
dashpole, line 156: should we be using the XCTC propagator here?

aabmass: Oops copypasta mistake thanks.

dashpole (approved):

PR #235, opened 2023-02-22
Refactor resource mapping logic into resource detection package

aabmass:
and update the exporters to use the shared code.

This means the trace and metric exporter now have a dependency on the resource detector package.


opentelemetry-exporter-gcp-trace/src/opentelemetry/exporter/cloud_trace/__init__.py
```diff
@@ -451,24 +440,18 @@ def _extract_resources(
...
-    resource_type = resource_attributes["gcp.resource_type"]
-    if (
-        not isinstance(resource_type, str)
-        or resource_type not in OT_RESOURCE_ATTRIBUTE_TO_GCP
+    monitored_resource = get_monitored_resource(resource)
+    # Do not map generic_task and generic_node to g.co/r/... span labels.
+    if monitored_resource and monitored_resource.type not in (
+        _resource_constants.GENERIC_NODE,
+        _resource_constants.GENERIC_TASK,
     ):
```
aabmass, line 448: Reason for this is: it doesn't seem like good UX to ever send the `g.co/r/generic*` since that is populated from service resource labels.

jsuereth (approved):

aabmass:
@dashpole lmk if you have any objections or I'll merge

dashpole (approved):

PR #244, opened 2023-05-11
Customize metric prefix

fengxumr:
By default, the metric prefix is "workload.googleapis.com". This change enables users to specific their own metric prefix.


aabmass (approved):
Thanks for the PR, LGTM

PR #256, opened 2023-07-05
Don't use `pkg_resources.get_distribution(..).version`

DouglasHeriot:
Similar to discussion at googleapis/python-api-core#27 I am using opentelemetry in an environment where modules are vendored without pip. So, pkg_resources can not be used to search for a module at runtime to get its version.

### DIFF
--- a/opentelemetry-exporter-gcp-monitoring/src/opentelemetry/exporter/cloud_monitoring/__init__.py
+++ b/opentelemetry-exporter-gcp-monitoring/src/opentelemetry/exporter/cloud_monitoring/__init__.py
@@ -19,7 +19,6 @@ from time import time_ns
 from typing import Dict, List, NoReturn, Optional, Set, Union
 
 import google.auth
-import pkg_resources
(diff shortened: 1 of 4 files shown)


aabmass (commented):
Can you also update here? https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/blob/8f8d2f0044bac36117089bd881ab902c330a9220/opentelemetry-exporter-gcp-trace/tests/test_cloud_trace_exporter.py#L98

DouglasHeriot:
Thanks for noticing that. I also found I missed another use in `opentelemetry-exporter-gcp-trace/src/opentelemetry/exporter/cloud_trace/__init__.py`.

DouglasHeriot:
This won’t be able to pass the tests until #270 (fix pinning to older `opentelemetry-sdk` version) and/or #269 (update `opentelemetry-sdk` version) are merged.

aabmass (approved):

PR #270, opened 2023-08-03
Pin upstream version of opentelemetry when running tests

DouglasHeriot:
The fix in #254 didn't apply to tox environments that run `pip install .`

In #256 `test_detects_gce` was failing with opentelemetry-sdk version 1.19.0 instead of 1.18.0

Also fix `get_mock_server.sh` not being in `allowlist_externals`


aabmass:
Thanks for the fix :)

Issue #272, opened 2023-08-22
Accept faas.name and faas.id for mapping metrics to generic_task


dashpole:
In order to map to `generic_task` on cloud run, cloud functions, and GAE, we need to map from `faas.name` and `faas.id` to `generic_task`, similar to what we do for `service.name` and `service.instance.id`

Example in golang:

https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/ef9b15e0c12cee086ab8a6e7439e64258b029a3b/internal/resourcemapping/resourcemapping.go#L228-L235

Current implementation:

https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/blob/b0ca7decb6a5bb01409822e746b7463f4a7a76ba/opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_mapping.py#L168-L173

PR #273, opened 2023-08-22
Map faas.* attributes to generic_task in resource mapping

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/issues/272

Question: should faas.* attributes take precedence over service.* attributes if they are present?  If they do, it might break users who are currently using service.* attributes.  If we don't, the default service (unknown) will take precedence over faas.name

### DIFF
--- a/opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_mapping.py
+++ b/opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_mapping.py
@@ -173,11 +173,8 @@ def get_monitored_resource(
     else:
         # fallback to generic_task
         if (
-            ResourceAttributes.SERVICE_NAME in attrs
-            and ResourceAttributes.SERVICE_INSTANCE_ID in attrs
-        ) or (
-            ResourceAttributes.FAAS_NAME in attrs
-            and ResourceAttributes.FAAS_INSTANCE in attrs
+            (ResourceAttributes.SERVICE_NAME in attrs or ResourceAttributes.FAAS_NAME in attrs)
+            and (ResourceAttributes.SERVICE_INSTANCE_ID in attrs or ResourceAttributes.FAAS_INSTANCE in attrs)
         ):
             mr = _create_monitored_resource(_constants.GENERIC_TASK, attrs)
         else:
@@ -194,24 +191,14 @@ def _create_monitored_resource(
 
     for mr_key, map_config in mapping.items():
         mr_value = None
-        fallback_value = None
         for otel_key in map_config.otel_keys:
-            if otel_key in resource_attrs:
-                if (
-                    otel_key == ResourceAttributes.SERVICE_NAME
-                    and isinstance(resource_attrs[otel_key], str)
-                    and str(resource_attrs[otel_key]).startswith(
-                        _constants.UNKNOWN_SERVICE_PREFIX
-                    )
-                ):
-                    # Only use values with an unknown_service prefix as the last resort
-                    fallback_value = resource_attrs[otel_key]
-                else:
-                    mr_value = resource_attrs[otel_key]
-                    break
+            if otel_key in resource_attrs and not str(resource_attrs[otel_key]).startswith(_constants.UNKNOWN_SERVICE_PREFIX):
+                mr_value = resource_attrs[otel_key]
+                break
 
-        if mr_value is None:
-            mr_value = fallback_value
+        if mr_value is None and ResourceAttributes.SERVICE_NAME in map_config.otel_keys:
+            # The service name started with unknown_service, and was ignored above.
+            mr_value = resource_attrs[ResourceAttributes.SERVICE_NAME]
 
         if mr_value is None:
             mr_value = map_config.fallback
--- a/opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_mapping.py
+++ b/opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_mapping.py
@@ -206,7 +206,7 @@ def _create_monitored_resource(
             and ResourceAttributes.SERVICE_NAME in map_config.otel_keys
         ):
             # The service name started with unknown_service, and was ignored above.
-            mr_value = resource_attrs[ResourceAttributes.SERVICE_NAME]
+            mr_value = resource_attrs.get(ResourceAttributes.SERVICE_NAME)
 
         if mr_value is None:
             mr_value = map_config.fallback
--- a/opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_mapping.py
+++ b/opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_mapping.py
@@ -110,8 +110,14 @@ MAPPINGS = {
             fallback="global",
         ),
         _constants.NAMESPACE: MapConfig(ResourceAttributes.SERVICE_NAMESPACE),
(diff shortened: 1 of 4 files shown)


dashpole:
Ideally: If a user sets a service.name, use that.  Otherwise, fall back to faas.name.

Options:

* resource detector detects service.name as faas.name.  (would this cause schema URL conflicts?)
* Check for serivce.name first, then fall back to faas.name.  But ignore the unknown service name.
* Faas.* takes precedence over service.name (this PR)

dashpole:
I've proposed option 2 in a design document.

dashpole:
I think this is good to go @aabmass.

opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_mapping.py
```diff
@@ -185,10 +194,24 @@ def _create_monitored_resource(
...
         for otel_key in map_config.otel_keys:
             if otel_key in resource_attrs:
-                mr_value = resource_attrs[otel_key]
-                break
+                if (
+                    otel_key == ResourceAttributes.SERVICE_NAME
+                    and isinstance(resource_attrs[otel_key], str)
+                    and str(resource_attrs[otel_key]).startswith(
+                        _constants.UNKNOWN_SERVICE_PREFIX
+                    )
```
aabmass, line 205: You should be able to do just one of these. Either assign to a local var before doing the isinstance check or just rely on `str()` convert whatever it is to a string.

opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_mapping.py
```diff
@@ -185,10 +194,24 @@ def _create_monitored_resource(
 
     for mr_key, map_config in mapping.items():
         mr_value = None
+        fallback_value = None
         for otel_key in map_config.otel_keys:
             if otel_key in resource_attrs:
-                mr_value = resource_attrs[otel_key]
-                break
+                if (
```
aabmass, line 200: Do you think it would be cleaner to move this out of the loop and do lazily directly at the L213 check? Either check by `if mr_key == _constants.TASK_ID and ...` or maybe `if mr_value.startswith(_constants.UNKNOWN_SERVICE_PREFIX)`.

opentelemetry-resourcedetector-gcp/tests/test_mapping.py
aabmass, line 1: Can you add a test case for the `unkown_service:foo` fallback?

opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_mapping.py
dashpole: done

opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_mapping.py
dashpole: moved.  It does look cleaner.

opentelemetry-resourcedetector-gcp/tests/test_mapping.py
dashpole: added

opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_mapping.py
```diff
@@ -186,10 +195,19 @@ def _create_monitored_resource(
...
         mr_value = None
         for otel_key in map_config.otel_keys:
-            if otel_key in resource_attrs:
+            if otel_key in resource_attrs and not str(
+                resource_attrs[otel_key]
+            ).startswith(_constants.UNKNOWN_SERVICE_PREFIX):
                 mr_value = resource_attrs[otel_key]
                 break
 
+        if (
```
aabmass, line 204: Won't this overwrite any MR label with service.name if it's None? For example location won't become "global" if there is a service.name set

opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_mapping.py
```diff
@@ -186,10 +195,19 @@ def _create_monitored_resource(
...
+            mr_value is None
+            and ResourceAttributes.SERVICE_NAME in map_config.otel_keys
+        ):
+            # The service name started with unknown_service, and was ignored above.
+            mr_value = resource_attrs[ResourceAttributes.SERVICE_NAME]
```
aabmass, line 209: Dict access will throw a KeyError for missing keys. Either check if the dict has the key or use `dict.get()`

aabmass: I'm not sure why there is no test failing for this, maybe I'm missing something

opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_mapping.py
dashpole: The next line, `and ResourceAttributes.SERVICE_NAME in map_config.otel_keys`, should mean that it only overwrites MR labels with service.name if service.name is one of the keys that was one of the fallbacks for the MR label.

opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_mapping.py
dashpole: No test is failing because all tests seem to use the fallback for service.name.  I'm not actually sure how to disable that behavior for tests...

dashpole: updated to use get()

aabmass: Do you mean because it always adds `service.name`? If you use the constructor directly `Resource({"foo": "bar"})`, it won't auto-add any OTel resource default stuff

```py
In [12]: Resource.create().attributes
Out[12]: BoundedAttributes({'telemetry.sdk.language': 'python', 'telemetry.sdk.name': 'opentelemetry', 'telemetry.sdk.version': '1.20.0', 'service.name': 'unknown_service'}, maxlen=None)

In [13]: Resource({}).attributes
Out[13]: BoundedAttributes({}, maxlen=None)
```

opentelemetry-resourcedetector-gcp/tests/test_mapping.py
aabmass: I didn't see a test case for this, just for the new FAAS stuff

dashpole: The `generic task faas fallback` case should test this, as it falls back to unknown_service.

opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_mapping.py
dashpole: Switched to creating resource with Resource({})

aabmass (approved):
Thanks!

opentelemetry-resourcedetector-gcp/tests/test_mapping.py
aabmass: Ooops i missed that, thanks

PR #302, opened 2023-11-22
chore: run codeql on PRs

dashpole:
Based on https://github.com/open-telemetry/opentelemetry-go/blob/main/.github/workflows/codeql-analysis.yml and https://github.com/open-telemetry/opentelemetry-python/blob/main/.github/workflows/codeql-analysis.yml


aabmass (approved):
Sorry, I never saw this!

PR #316, opened 2024-04-30
Add grpc trace export sample

psx95:
The added gRPC sample also showcases credential refresh process.

#### Testing
 - `python3 example_grpc.py`
 -  Verified that expired credentials were refreshing through added logs.


samples/otlptrace/README.md
```diff
@@ -0,0 +1,32 @@
...
+#### Run the Sample
+```sh
+# export necessary OTEL environment variables
+export OTEL_RESOURCE_ATTRIBUTES="gcp.project_id=<project-id>"
+export OTEL_EXPORTER_OTLP_ENDPOINT=<endpoint>
+
+cd samples/otlptrace && python3 example-grpc.py
+```
+Other variations of the sample:
+ - `python3 example.py` - will run a program that will export traces using http/protobuf.
```
aabmass, line 32: Is the script you renamed in the other PR?

samples/otlptrace/example-grpc.py
aabmass, line 1: Same comment about underscores

aabmass, line 1: Can you run `black <filename>` and `isort <filename>` on these files to auto format (the CI doesn't pick up this directory)? You can create a repo level virtual env with the correct versions of those programs with `tox -e dev`.

samples/otlptrace/example-grpc.py
```diff
@@ -0,0 +1,82 @@
...
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import BatchSpanProcessor
+import grpc
+import logging
+import time
+
+"""
+This is a sample script that exports OTLP traces encoded as protobufs via gRPC. 
+"""
+class AutoRefreshAuthMetadataPlugin(AuthMetadataPlugin):
```
aabmass, line 34: Is it not sufficient to use the parent class directly? I thought it was for this purpose specifically

aabmass (commented):
Can you run `black` and `isort` tools on these files?

samples/otlptrace/README.md
psx95: Nope, this needs to be updated.

samples/otlptrace/example-grpc.py
psx95: No, the default implementation seems to add empty headers to the request

```py
    def __init__(self, credentials, request, default_host=None):        
        super(AuthMetadataPlugin, self).__init__()
        self._credentials = credentials
        self._request = request
        self._default_host = default_host

    def _get_authorization_headers(self, context):
        headers = {}

        # https://google.aip.dev/auth/4111
        # Attempt to use self-signed JWTs when a service account is used.
        # A default host must be explicitly provided since it cannot always
        # be determined from the context.service_url.
        if isinstance(self._credentials, service_account.Credentials):
            self._credentials._create_self_signed_jwt(
                "https://{}/".format(self._default_host) if self._default_host else None
            )

        self._credentials.before_request(
            self._request, context.method_name, context.service_url, headers
        )

        return list(six.iteritems(headers))

    def __call__(self, context, callback):
        callback(self._get_authorization_headers(context), None)
```

I also tried it directly and it failed to export once the credentials expired.

samples/otlptrace/example-grpc.py
psx95: Done, is there a way we can add the directory to the CI ? Looked at the actions file, did not find an obvious place to add this directory at any place.

samples/otlptrace/example_grpc.py
```diff
@@ -0,0 +1,97 @@
...
+        auth_headers = [("authorization", f"Bearer {self._credentials.token}")]
+        callback(auth_headers, None)
+
+
+credentials, project_id = google.auth.default()
+request = google.auth.transport.requests.Request()
+credentials.refresh(request)
+resource = Resource.create(attributes={SERVICE_NAME: "otlp-gcp-grpc-sample"})
+
+auth_metadata_plugin = AutoRefreshAuthMetadataPlugin(
```
aabmass, line 58: (Continuing the other thread here)

I was able to get it working with `AuthMetadataPlugin` directly and it ran successfully overnight.
```diff
diff --git a/samples/otlptrace/example_grpc.py b/samples/otlptrace/example_grpc.py                                                           
index cb53cd1..f8d1820 100644
--- a/samples/otlptrace/example_grpc.py                                                                                                      
+++ b/samples/otlptrace/example_grpc.py
@@ -55,7 +55,7 @@ request = google.auth.transport.requests.Request()                                                                         
 credentials.refresh(request)
 resource = Resource.create(attributes={SERVICE_NAME: "otlp-gcp-grpc-sample"})
 
-auth_metadata_plugin = AutoRefreshAuthMetadataPlugin(
+auth_metadata_plugin = AuthMetadataPlugin(                                                                                                  
     credentials=credentials, request=request
 )
 channel_creds = grpc.composite_channel_credentials(
@@ -68,9 +68,6 @@ processor = BatchSpanProcessor(
     OTLPSpanExporter(
         credentials=channel_creds,
         insecure=False,
-        headers={
-            "x-goog-user-project": credentials.quota_project_id,
-        },
     )
 )
 trace_provider.add_span_processor(processor)
```

The problem was L72 adds `x-goog-user-project` but this header was also being added by the AuthMetadataPlugin causing duplicate headers to be comma joined and an `INVALID_ARGUMENT` response.

psx95: Good find! Thanks @aabmass !

samples/otlptrace/README.md
```diff
@@ -0,0 +1,32 @@
...
+gcloud auth application-default login
+```
+
+#### Run the Sample
+```sh
+# export necessary OTEL environment variables
+export OTEL_RESOURCE_ATTRIBUTES="gcp.project_id=<project-id>"
+export OTEL_EXPORTER_OTLP_ENDPOINT=<endpoint>
+
+cd samples/otlptrace && python3 example_grpc.py
```
aabmass, line 29: nit the `cd` is mentioned above so might be confusing/repetitive here

samples/otlptrace/example_grpc.py
```diff
@@ -0,0 +1,74 @@
...
+from opentelemetry.sdk.trace.export import BatchSpanProcessor
+
+"""
+This is a sample script that exports OTLP traces encoded as protobufs via gRPC. 
+"""
+
+
+credentials, project_id = google.auth.default()
+request = google.auth.transport.requests.Request()
+credentials.refresh(request)
```
aabmass, line 38: nit this seems "eager" but I think it would lazily happen otherwise. Can we remove?

aabmass (approved):

samples/otlptrace/example_grpc.py
```diff
@@ -0,0 +1,74 @@
...
+    credentials=credentials, request=request
+)
+channel_creds = grpc.composite_channel_credentials(
+    grpc.ssl_channel_credentials(),
+    grpc.metadata_call_credentials(auth_metadata_plugin),
+)
+
+trace_provider = TracerProvider(resource=resource)
+processor = BatchSpanProcessor(
+    OTLPSpanExporter(credentials=channel_creds, insecure=False)
```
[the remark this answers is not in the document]
psx95: Based on the control flow, in `OTLPExporterMixin`: 

```py
 if insecure:
            self._client = self._stub(
                insecure_channel(self._endpoint, compression=compression)
            )
        else:
            credentials = _get_credentials(
                credentials, OTEL_EXPORTER_OTLP_CERTIFICATE
            )
            self._client = self._stub(
                secure_channel(
                    self._endpoint, credentials, compression=compression
                )
            )
```

Makes me think that in an insecure channel, credentials would not be passed. However, I did not try running the example with `insecure=True`

samples/otlptrace/README.md
psx95: I will create a follow-up PR to address these, I had auto-merge on so it merged without waiting for comments to be resolved.

samples/otlptrace/example_grpc.py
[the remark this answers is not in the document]
aabmass: Gotcha. But can we instead rely on the default behavior and just remove the argument altogether?

samples/otlptrace/README.md
aabmass: Sorry about that diidn't realize

samples/otlptrace/example_grpc.py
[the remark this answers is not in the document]
psx95: I see - it seems like the default is `None` but if the parsed url scheme is https, it is automatically assigned false.

This makes me think that it depends on the endpoint. If the endpoint uses `http` instead of `https`, then without explicitly defining `insecure=False`, the code might fail.

Does this make sense ?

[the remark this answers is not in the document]
aabmass: Yes I think it should behave correctly

psx95:
Addressed all minor feedback in #321.

PR #317, opened 2024-05-01
Rename example.py -> example-http.py

psx95:


samples/otlptrace/example-http.py
aabmass, line 1: Should use underscores for filenames in python

aabmass (changes requested):

aabmass (approved):

PR #326, opened 2024-05-09
Add WIP instrumentation quickstart sample

dashpole:
This is mostly boilerplate to make future PRs smaller.

The following are taken from https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/pull/309:

* `samples/instrumentation-quickstart/Dockerfile`
* `samples/instrumentation-quickstart/app.py`

The following are minimally adapted from the java quickstart:

* samples/instrumentation-quickstart/README.md
* samples/instrumentation-quickstart/docker-compose.creds.yaml
* samples/instrumentation-quickstart/docker-compose.yaml

The quickstart runs and serves requests, but does not currently include any instrumentation.  The logs are also not in the correct format.

~~I noticed we have some examples in docs/examples, and another example in /samples.  Should we consolidate?~~

### DIFF
--- a/samples/instrumentation-quickstart/Dockerfile
+++ b/samples/instrumentation-quickstart/Dockerfile
@@ -3,4 +3,6 @@ WORKDIR /usr/src/app
 COPY . .
 RUN python -m venv /venv && \
     /venv/bin/pip install -r requirements.txt
+# TODO: Make this a more realistic productionized application using Gunicorn as a SWGI server.
+# See https://flask.palletsprojects.com/en/3.0.x/deploying/ 
 CMD . /venv/bin/activate && flask run --host 0.0.0.0 --port 8080 2>&1 | tee /var/log/app.log
--- a/samples/instrumentation-quickstart/Dockerfile
+++ b/samples/instrumentation-quickstart/Dockerfile
@@ -3,4 +3,4 @@ WORKDIR /usr/src/app
 COPY . .
 RUN python -m venv /venv && \
     /venv/bin/pip install -r requirements.txt
-CMD bash -c "source /venv/bin/activate && flask run --host 0.0.0.0 --port 8080 2>&1 | tee /var/log/app.log"
+CMD . /venv/bin/activate && flask run --host 0.0.0.0 --port 8080 2>&1 | tee /var/log/app.log
--- a/samples/instrumentation-quickstart/Dockerfile
+++ b/samples/instrumentation-quickstart/Dockerfile
@@ -1,0 +1,6 @@
+FROM python:3.12
+WORKDIR /usr/src/app
+COPY . .
+RUN python -m venv /venv && \
+    /venv/bin/pip install -r requirements.txt
+CMD bash -c "source /venv/bin/activate && flask run --host 0.0.0.0 --port 8080 2>&1 | tee /var/log/app.log"
--- a/samples/instrumentation-quickstart/README.md
+++ b/samples/instrumentation-quickstart/README.md
@@ -29,7 +29,7 @@ following roles to run the example:
 
 ```sh
 git clone https://github.com/GoogleCloudPlatform/opentelemetry-operations-python.git
-cd opentelemetry-operations-python/examples/instrumentation-quickstart
+cd opentelemetry-operations-python/samples/instrumentation-quickstart
 docker compose up --abort-on-container-exit
 ```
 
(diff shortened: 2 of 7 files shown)


aabmass:
> I noticed we have some examples in docs/examples, and another example in /samples. Should we consolidate?

The docs under docs/examples are for https://google-cloud-opentelemetry.readthedocs.io/en/latest/. Let just leave it in /samples for now

dashpole:
I'm going to switch to manual instrumentation

dashpole:
I think this is ready for review.  The app builds and runs successfully, but it does not include any instrumentation.  I'll do that as follow-ups

dashpole:
@aabmass any recommendation on testing this as part of CI?  Should I use cloud build similar to the java quickstart?

samples/instrumentation-quickstart/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM python:3.12
+WORKDIR /usr/src/app
+COPY . .
+RUN python -m venv /venv && \
+    /venv/bin/pip install -r requirements.txt
+CMD bash -c "source /venv/bin/activate && flask run --host 0.0.0.0 --port 8080 2>&1 | tee /var/log/app.log"
```
aabmass, line 6: When i was working on the [JS sample](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/v2.2.0/v0.18.0/samples/instrumentation-quickstart/Dockerfile), I realized the `bash -c` is not actually needed since docker uses a real shell for CMD

aabmass, line 6: I think for this sample it would be nice to have a more realistic productionized app. Typically people use Gunicorn as a WSGI server, see https://flask.palletsprojects.com/en/3.0.x/deploying/.

Feel free to add a TODO or address in another PR

samples/instrumentation-quickstart/README.md
```diff
@@ -0,0 +1,88 @@
...
+- `roles/monitoring.metricWriter` – see https://cloud.google.com/monitoring/access-control#predefined_roles
+- `roles/cloudtrace.agent` – see https://cloud.google.com/trace/docs/iam#trace-roles
+
+## Running the example
+
+### Cloud Shell or GCE
+
+```sh
+git clone https://github.com/GoogleCloudPlatform/opentelemetry-operations-python.git
+cd opentelemetry-operations-python/examples/instrumentation-quickstart
```
aabmass, line 32:
```suggestion
cd opentelemetry-operations-python/samples/instrumentation-quickstart
```

samples/instrumentation-quickstart/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+Learn more at [Setting Up Authentication for Server to Server Production Applications][ADC].
+
+*Note:* Application Default Credentials is able to implicitly find the credentials as long as the application is running on Compute Engine, Kubernetes Engine, App Engine, or Cloud Functions.
+
+Then run the example:
+
+```sh
+git clone https://github.com/GoogleCloudPlatform/opentelemetry-operations-python.git
+cd opentelemetry-operations-python/examples/instrumentation-quickstart
```
aabmass, line 72:
```suggestion
cd opentelemetry-operations-python/samples/instrumentation-quickstart
```

samples/instrumentation-quickstart/app.py
```diff
@@ -0,0 +1,37 @@
...
+    # TODO: add info log line here
+    for _ in range(randint(3, 7)):
+        requests.get(url_for('single', _external=True))
+    return 'ok'
+
+@app.route('/single')
+def single():
+    duration = uniform(0.1, 0.2)
+    time.sleep(duration)
+    return 'slept ' + str(duration) + ' seconds'
```
aabmass, line 37: nit prefer f-string
```suggestion
    return f'slept {duration} seconds'
```

samples/instrumentation-quickstart/app.py
```diff
@@ -0,0 +1,37 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from random import randint, uniform
+from flask import Flask, url_for
+import requests
+import time
+
+app = Flask(__name__, static_folder=None)
```
aabmass, line 20: Was `static_folder=None` needed?

aabmass (approved):

samples/instrumentation-quickstart/app.py
dashpole: doesn't seem like it.  removed

samples/instrumentation-quickstart/Dockerfile
dashpole: Removed.  I had to change "source" to "." because of the change from `bash` to `sh`.

dashpole: added a TODO

PR #328, opened 2024-05-10
Add json logging to the sample application

dashpole:
This uses `python-json-logger` to do the json formatting, and changes the keys used to match what GCP expects.  All loglevels seem to work properly.  I did not investigate other logging libraries, since the existing GCP docs already recommend using `logging`.

Screenshot 2024-05-10 at 2 43 53 PM

The second commit adds the trace context keys.  I used `opentelemetry-instrumentation-logging` to make the keys available to the logger, and used `rename_fields` to set the proper GCP keys for the trace context.

### DIFF
--- a/samples/instrumentation-quickstart/app.py
+++ b/samples/instrumentation-quickstart/app.py
@@ -15,6 +15,9 @@
 from random import randint, uniform
 import time
 
+import logging
+from pythonjsonlogger import jsonlogger
+
 from opentelemetry import trace
 from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
 from opentelemetry.sdk.trace import TracerProvider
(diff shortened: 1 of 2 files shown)


samples/instrumentation-quickstart/app.py
```diff
@@ -31,6 +35,27 @@
...
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+logHandler = logging.StreamHandler()
+formatter = jsonlogger.JsonFormatter(
+    "%(asctime)s %(levelname)s %(message)s %(otelTraceID)s %(otelSpanID)s %(otelTraceSampled)s",
+    rename_fields={
+        "levelname": "severity",
+        "asctime": "timestamp",
+        "otelTraceID": "logging.googleapis.com/trace",
+        "otelSpanID": "logging.googleapis.com/spanId",
```
quentinmit, line 49: Isn't there an OTel spec for trace and span ID that doesn't match this?

I understand what you have here works with most GCP agents, but it seems suboptimal to have an OTel sample use a legacy GCP-specific key when there's a generic alternative.

aabmass (approved):

samples/instrumentation-quickstart/app.py
dashpole: Yes.  The spec is currently experimental: https://opentelemetry.io/docs/specs/otel/compatibility/logging_trace_context/.  We are separately working on getting those keys supported by agents.  Once that support exists, we will update the samples to use those keys.

PR #329, opened 2024-05-13
Change the sample to use gunicorn

dashpole:
Changes:

* Use gunicorn
* Manually configure the resource in the app instead of using the OTEL_SERVICE_NAME env var.
* Use the PID as the service.instance.id to avoid duplicate timeseries errors.
* Change the `resource` processor in the collector to provide defaults, and not to override the resource from the application

### DIFF
--- a/samples/instrumentation-quickstart/Dockerfile
+++ b/samples/instrumentation-quickstart/Dockerfile
@@ -3,4 +3,4 @@ WORKDIR /usr/src/app
 COPY . .
 RUN python -m venv /venv && \
     /venv/bin/pip install -r requirements.txt
-CMD . /venv/bin/activate && gunicorn -b 0.0.0.0:8080 -w 4 'app:app'  2>&1 | tee /var/log/app.log
+CMD /venv/bin/gunicorn -b 0.0.0.0:8080 -w 4 'app:app'  2>&1 | tee /var/log/app.log
--- a/samples/instrumentation-quickstart/Dockerfile
+++ b/samples/instrumentation-quickstart/Dockerfile
@@ -3,6 +3,4 @@ WORKDIR /usr/src/app
 COPY . .
 RUN python -m venv /venv && \
     /venv/bin/pip install -r requirements.txt
-# TODO: Make this a more realistic productionized application using Gunicorn as a SWGI server.
-# See https://flask.palletsprojects.com/en/3.0.x/deploying/ 
-CMD . /venv/bin/activate && flask run --host 0.0.0.0 --port 8080 2>&1 | tee /var/log/app.log
+CMD . /venv/bin/activate && gunicorn -b 0.0.0.0:8080 -w 4 'app:app'  2>&1 | tee /var/log/app.log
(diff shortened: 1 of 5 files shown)


samples/instrumentation-quickstart/otel-collector-config.yaml
```diff
@@ -98,14 +98,14 @@ processors:
...
   resource:
     attributes:
-      - { key: "location", value: "us-central1", action: "upsert" }
-      - { key: "cluster", value: "no-cluster", action: "upsert" }
-      - { key: "namespace", value: "no-namespace", action: "upsert" }
-      - { key: "job", value: "us-job", action: "upsert" }
-      - { key: "instance", value: "us-instance", action: "upsert" }
+      - { key: "cloud.region", value: "us-central1", action: "insert" }
+      - { key: "k8s.cluster.name", value: "no-cluster", action: "insert" }
+      - { key: "k8s.namespace.name", value: "no-namespace", action: "insert" }
```
aabmass, line 106: Are the `k8s.*` keys preferred over the ones documented here https://cloud.google.com/stackdriver/docs/managed-prometheus/setup-otel#detect-resource-attributes?

samples/instrumentation-quickstart/Dockerfile
```diff
@@ -3,6 +3,4 @@ WORKDIR /usr/src/app
 COPY . .
 RUN python -m venv /venv && \
     /venv/bin/pip install -r requirements.txt
-# TODO: Make this a more realistic productionized application using Gunicorn as a SWGI server.
-# See https://flask.palletsprojects.com/en/3.0.x/deploying/ 
-CMD . /venv/bin/activate && flask run --host 0.0.0.0 --port 8080 2>&1 | tee /var/log/app.log
\ No newline at end of file
+CMD . /venv/bin/activate && gunicorn -b 0.0.0.0:8080 -w 4 'app:app'  2>&1 | tee /var/log/app.log
```
aabmass, line 6: Does this work?
```suggestion
CMD /venv/bin/gunicorn -b 0.0.0.0:8080 -w 4 'app:app'  2>&1 | tee /var/log/app.log
```

samples/instrumentation-quickstart/otel-collector-config.yaml
dashpole: I wanted these to be defaults, and not to override what users set/detect.  If we use the ones from that document, it will override anything from the user (e.g. `job` takes precedence over `service.name`).

samples/instrumentation-quickstart/docker-compose.yaml
```diff
@@ -19,7 +19,6 @@ services:
     build: .
     environment:
       - OTEL_EXPORTER_OTLP_ENDPOINT=http://otelcol:4318
-      - OTEL_SERVICE_NAME=otel-quickstart-python
```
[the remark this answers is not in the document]
dashpole: yep, that works!

samples/instrumentation-quickstart/Dockerfile
dashpole: yep, works!

samples/instrumentation-quickstart/otel-collector-config.yaml
aabmass: Are users expected to set `job` in their OTel resource?

aabmass (approved):

samples/instrumentation-quickstart/otel-collector-config.yaml
dashpole: They are not expected to set job.

PR #331, opened 2024-05-15
Clean up sample, and add tags

dashpole:
Changes:

* Split logging and OTel setup into their own files
* remove unneeded `logging.getLogger('werkzeug').setLevel(logging.ERROR)` line.
* Use `logging.basicConfig(` to configure the handler globally, instead of only configuring the handler in the app file.
* Add tags so this can be used in GCP documentation

### DIFF
--- a/samples/instrumentation-quickstart/app.py
+++ b/samples/instrumentation-quickstart/app.py
@@ -13,73 +13,25 @@
 # limitations under the License.
 
 from random import randint, uniform
-import os
 import time
-
 import logging
-from pythonjsonlogger import jsonlogger
-from opentelemetry.instrumentation.logging import LoggingInstrumentor
-
-from opentelemetry.sdk.resources import SERVICE_INSTANCE_ID, SERVICE_NAME, Resource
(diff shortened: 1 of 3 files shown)


samples/instrumentation-quickstart/app.py
```diff
@@ -13,83 +13,38 @@
...
+logger = logging.getLogger(__name__)
 
 app = Flask(__name__)
 FlaskInstrumentor().instrument_app(app)
 RequestsInstrumentor().instrument()
+# [END opentelemetry_instrumentation_main]
 
+# [START opentelemetry_instrumentation_handle_multi]
 @app.route('/multi')
 def multi():
```
aabmass, line 36: We have a docstring in the other lang examples for each route. Can you copy it for this sample too?

aabmass (approved):

samples/instrumentation-quickstart/app.py
dashpole: added

PR #339, opened 2024-07-17
Use envion to get the values

arnavmehta7:


arnavmehta7:
Hey, I don't think so there exists an issue. I just needed the configuration and then raised the PR.

PR #340, opened 2024-07-18
Generate updated requirements.txt for examples using pip-compile

aabmass:
`pip-compile` is a tool to generate requirements.txt file with locked transitive dependencies from a requirements.in containing only immediate dependencies https://github.com/jazzband/pip-tools/.

This PR adds `requirements.in` files for the examples and `pip-compile`s them. This should resolve all the dependenabot "alerts" and PRs since the dependencies will be up to date. Renovate (or dependabot maybe) should be able to do this automatically in the future.

I manually tested the examples and they are working with the upgraded deps.


dashpole (approved):
Do we need to update CI to make sure pip-compile is run during presubmit?

aabmass:
> Do we need to update CI to make sure pip-compile is run during presubmit?

I don't think we should run it during presubmit. Renovate/dependabot can send updates in separate PRs I believe.

PR #342, opened 2024-07-18
Update a few examples broken or misleading examples

aabmass:
- Remove any usage of SimpleSpanProcessor
- Fix import paths for resource detector which were missed earlier
- Fix use of pre-1.0 metrics SDK


dashpole (commented):
do you need to add any flush handling when switching to the batch processor?

aabmass:
OTel adds an [`atexit`](https://docs.python.org/3/library/atexit.html) handler automatically which will run before the python process exits normally (which these examples do). It's on by default https://github.com/open-telemetry/opentelemetry-python/blob/b1e99c1555721f818e578d7457587693e767e182/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L1195

dashpole (approved):

PR #345, opened 2024-07-22
Small fixups for resource detector code and tests

aabmass:
- Updated the tests to not leak into os.environ
- Fix code to expect numeric value for `instance/id` metadata
- Mark metadata response JSON type annotation for `instance.attributes` as not-total (keys are not all required)


opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_gce.py
```diff
@@ -45,7 +45,7 @@ def host_type() -> str:
 
 
 def host_id() -> str:
-    return _metadata.get_metadata()["instance"]["id"]
+    return str(_metadata.get_metadata()["instance"]["id"])
```
xrmx, line 48: I don't have a google cloud instance at hand but are you sure this is not already a string?

aabmass: Yes it's a JSON int at least some of the time. It might return a string if it's too big for a JS integer
```
curl -s -H 'Content-Type: text/json' -H 'Metadata-Flavor:Google'  metadata/computeMetadata/v1/?recursive=true | jq '.instance.id'
2459451723172637643
```

xrmx: thanks for double checking

xrmx (approved):

dashpole:
is instance/id usually an int?  I thought it was a UID on cloud run...

aabmass:
> is instance/id usually an int? I thought it was a UID on cloud run...

Hmm that's a good point. On pGCE it's numeric](https://cloud.google.com/compute/docs/metadata/predefined-metadata-keys#:~:text=The%20ID%20of%20the%20VM.%20This%20is%20a%20unique%2C%20numerical%20ID%20that%20is%20generated%20by%20Compute%20Engine.%20This%20is%20useful%20for%20identifying%20VMs%20if%20you%20don%27t%20use%20VM%20names.). Let me see if I can just treat it as an opaque string I don't remember why I changed it anymore

opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_metadata.py
```diff
@@ -34,13 +35,14 @@ class Project(TypedDict):
...
 Attributes = TypedDict(
-    "Attributes", {"cluster-location": str, "cluster-name": str}
+    "Attributes", {"cluster-location": str, "cluster-name": str}, total=False
 )
 
 
 class Instance(TypedDict):
     attributes: Attributes
-    id: str
+    # id can be an integer on GCE VMs or a string on other environments
```
aabmass, line 44: @dashpole updated, does this address your comment?

dashpole: yep

dashpole (approved):

aabmass:
The "ops-python-e2e-cloud-functions-gen2 (opentelemetry-ops-e2e)" CI failure is because I enabled the checked but implementation is in a draft PR still. Merging

PR #362, opened 2024-12-06
Use a shorter timeout for the metadata endpoint

xrmx:
Use a shorter timeout of 2 seconds for the connection before the 5 seconds read one to avoid blocking too much if the metadata server is not available.

This makes the code behave more similar to the go one: https://github.com/googleapis/google-cloud-go/blob/main/compute/metadata/metadata.go#L65

### DIFF
--- a/opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_metadata.py
+++ b/opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_metadata.py
@@ -25,7 +25,8 @@ _GCP_METADATA_URL = "http://metadata.google.internal/computeMetadata/v1/"
 _INSTANCE = "instance"
 _RECURSIVE_PARAMS = {"recursive": "true"}
 _GCP_METADATA_URL_HEADER = {"Metadata-Flavor": "Google"}
-_TIMEOUT_SEC = 5
+# Use a shorter timeout for connection so we won't block much if it's unreachable
+_TIMEOUT = (2, 5)
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_metadata.py
```diff
@@ -25,7 +25,8 @@
 _INSTANCE = "instance"
 _RECURSIVE_PARAMS = {"recursive": "true"}
 _GCP_METADATA_URL_HEADER = {"Metadata-Flavor": "Google"}
-_TIMEOUT_SEC = 5
+# Use a shorter timeout for connection so we won't block much if it's unreachable
+_TIMEOUT = (2, 5)
```
aabmass, line 29: This is decreasing the connect timeout and not read timeout IIUC https://requests.readthedocs.io/en/latest/user/advanced/#timeouts?

Is that what you meant?

xrmx: Yes, the use case is reducing the timeout if loading the resource detector outside GCP

aabmass: Makes sense, can you just update the PR description s/read/connect?

aabmass (approved):

opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/_metadata.py
aabmass: Oh and also a changelog entry pls https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/blob/main/opentelemetry-resourcedetector-gcp/CHANGELOG.md

xrmx: Updated title, description and added CHANGELOG

PR #364, opened 2024-12-12
Update instrumentation sample to use exponential histograms

dashpole:
I'm trying to debug a customer issue with the python SDK + GMP exporter with exponential histograms, so i'm updating the sample to use exponential histograms.

I verified that exponential histogram  can be queried using both MQL and promql (using histogram_quantile).

### DIFF
--- a/samples/instrumentation-quickstart/docker-compose.yaml
+++ b/samples/instrumentation-quickstart/docker-compose.yaml
@@ -28,7 +28,7 @@ services:
     ports:
       - 8080
   otelcol:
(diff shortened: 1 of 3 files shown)


dashpole:
I think maybe the self-observability metrics broke.

aabmass:
@dashpole do we want to update all the languages to exponential?

dashpole:
Yes, I have a tracking bug open to do that

aabmass (approved):
🎉

PR #366, opened 2024-12-19
Fix creation of resources in _detector

### LINKED ISSUE #363: _detector.GoogleCloudResourceDetector misbehaving on Buildkite
I have `_detector.GoogleCloudResourceDetector` misbehaving when running under Buildkite (CI tool) on GKE.

When I load the resource detector the process will get restarted and I'll have an infinite loop with one of this messages from the OTel sdk per run:

```
Detector <GoogleCloudResourceDetector object at 0x7f8ae0522250> took longer than 5 seconds, skipping
```
I've debugged this a bit and there is something in `_gke_resource` that causes this.

As an experiment I've open coded the thing into another resource detector and to my surprise the following code is working fine:

```
class GoogleDebugCloudResourceDetector(ResourceDetector):
    def detect(self) -> Resource:
        from opentelemetry.resourcedetector.gcp_resource_detector import _metadata, _gke  # , _detector
        from opentelemetry.resourcedetector.gcp_resource_detector._constants import (
            ResourceAttributes,
        )

        try:
            print(_metadata.get_metadata())
        except Exception:
            return Resource.get_empty()

        if _gke.on_gke():
            cluster_location = _metadata.get_metadata()["instance"]["attributes"]["cluster-location"]
            hyphen_count = cluster_location.count("-")
            if hyphen_count == 1:
                zone_or_region_key = ResourceAttributes.CLOUD_REGION
            elif hyphen_count == 2:
                zone_or_region_key = ResourceAttributes.CLOUD_AVAILABILITY_ZONE
            else:
                print("oops no zone_or_region_key")
                zone_or_region_key = "oops"

            cluster_name = _metadata.get_metadata()["instance"]["attributes"]["cluster-name"]
            host_id = str(_metadata.get_metadata()["instance"]["id"])
            return Resource(
                {
                    ResourceAttributes.CLOUD_PLATFORM_KEY: ResourceAttributes.GCP_KUBERNETES_ENGINE,
                    zone_or_region_key: cluster_location,
                    ResourceAttributes.K8S_CLUSTER_NAME: cluster_name,
                    ResourceAttributes.HOST_ID: host_id,
                }
            )
```

Unrelated questions:
- WDYT on making `_metadata.is_available` a wrapper for _metadata.get_metadata() that returns false if an exception is raised? This way we have one http call less.

xrmx:
We should create a Resource instance and not use Resource.create because if we set OTEL_EXPERIMENTAL_RESOURCE_DETECTORS we will go into an infinite loop trying to load and instantiate all the resources detectors.

See [Resource.create implementation](https://github.com/open-telemetry/opentelemetry-python/blob/b48223561acfdffd1411ed7d015ddabe6f6eb801/opentelemetry-sdk/src/opentelemetry/sdk/resources/__init__.py#L172)

Fix #363

### DIFF
--- a/e2e-test-server/e2e_test_server/scenarios.py
+++ b/e2e-test-server/e2e_test_server/scenarios.py
@@ -29,6 +29,7 @@ from opentelemetry.sdk.trace import TracerProvider
 from opentelemetry.sdk.trace.export import BatchSpanProcessor
 from opentelemetry.sdk.trace.sampling import ALWAYS_ON
 from opentelemetry.trace import SpanKind, Tracer, format_trace_id
+from opentelemetry.sdk.resources import get_aggregated_resources
 from pydantic import BaseModel
 
 from .constants import INSTRUMENTING_MODULE_NAME, PROJECT_ID, TEST_ID, TRACE_ID
@@ -59,9 +60,7 @@ def _tracer_setup(
     spans created during the test after.
     """
 
-    tracer_provider = TracerProvider(
-        sampler=ALWAYS_ON, **tracer_provider_config
-    )
+    tracer_provider = TracerProvider(sampler=ALWAYS_ON, **tracer_provider_config)
     tracer_provider.add_span_processor(
         BatchSpanProcessor(
             CloudTraceSpanExporter(project_id=PROJECT_ID, **exporter_config)
@@ -139,9 +138,9 @@ def detect_resource(request: Request) -> Response:
     """Create a trace with GCP resource detector"""
     with _tracer_setup(
         tracer_provider_config={
-            "resource": GoogleCloudResourceDetector(
-                raise_on_error=True
-            ).detect()
+            "resource": get_aggregated_resources(
+                [GoogleCloudResourceDetector(raise_on_error=True)]
+            )
         },
         exporter_config={"resource_regex": r".*"},
     ) as tracer:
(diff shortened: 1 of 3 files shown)


aabmass:
We have  [discourage using the constructor](https://github.com/open-telemetry/opentelemetry-python/blob/b48223561acfdffd1411ed7d015ddabe6f6eb801/opentelemetry-sdk/src/opentelemetry/sdk/resources/__init__.py#L25-L26). If we only call the constructor, there is no way to get these attributes https://github.com/open-telemetry/opentelemetry-python/blob/b48223561acfdffd1411ed7d015ddabe6f6eb801/opentelemetry-sdk/src/opentelemetry/sdk/resources/__init__.py#L217-L227 or `OTEL_RESOURCE_ATTRIBUTES` based ones.

I think we might need to fix it this in the SDK or at least update the documentation to say that detectors MUST call the constructor. It seems a bit fragile overall, wdyt?

xrmx:
> We have [discourage using the constructor](https://github.com/open-telemetry/opentelemetry-python/blob/b48223561acfdffd1411ed7d015ddabe6f6eb801/opentelemetry-sdk/src/opentelemetry/sdk/resources/__init__.py#L25-L26). If we only call the constructor, there is no way to get these attributes https://github.com/open-telemetry/opentelemetry-python/blob/b48223561acfdffd1411ed7d015ddabe6f6eb801/opentelemetry-sdk/src/opentelemetry/sdk/resources/__init__.py#L217-L227 or `OTEL_RESOURCE_ATTRIBUTES` based ones.
> 
> I think we might need to fix it this in the SDK or at least update the documentation to say that detectors MUST call the constructor. It seems a bit fragile overall, wdyt?

Opened a [pr to update the docs](https://github.com/open-telemetry/opentelemetry-python/pull/4372)

aabmass:
@xrmx to fix the tests, can you regenerate the snapshots by running? That will highlight the difference in behavior too
```
tox -f py3xx-test -- --snapshot-update
```

aabmass (approved):

PR #371, opened 2025-01-17
feat: add otlp metric sample with gcp auth

dashpole:
TESTED=manually tested against staging


aabmass (approved):

dashpole:
Updated to add the gcp resource detector, which sets required fields (e.g. location).

PR #386, opened 2025-02-28
Update OTLP trace example.

DylanRussell:


samples/otlptrace/example_http.py
```diff
@@ -19,21 +19,22 @@
...
-
-trace_provider = TracerProvider(resource=resource)
-processor = BatchSpanProcessor(OTLPSpanExporter(headers=req_headers))
+trace_provider = TracerProvider(resource=Resource.create(attributes={SERVICE_NAME: "otlp-gcp-http-sample"}))
+processor = BatchSpanProcessor(
+    OTLPSpanExporter(
+        headers={
+            "x-goog-user-project": credentials.quota_project_id,
+        },
+        session=AuthorizedSession(credentials),
```
aabmass, line 34: Does `AuthorizedSession` already automatically add `x-goog-user-project` if needed?

DylanRussell: No. We could add the header to the AuthorizedSession object instead (it has a headers attribute that can be modified), but it doesn't automatically add it.

aabmass: Seems to be added on my system, it's possible you don't have a quota project set in your ADC

```console
$ nc -l 9999 | grep --invert-match 'Bearer'
GET / HTTP/1.1
Host: localhost:9999
User-Agent: python-requests/2.32.3
Accept-Encoding: gzip, deflate
Accept: */*
Connection: keep-alive
x-goog-api-client: cred-type/u
x-goog-user-project: my-project
```

```
uvx --with google-auth --with requests python -c \
  'import google.auth; from google.auth.transport.requests import AuthorizedSession; creds, _ = google.auth.default(); AuthorizedSession(creds).get("http://localhost:9999")'
```

DylanRussell: You're right: https://screenshot.googleplex.com/nuGUfiX5WBM3V5L

I didn't try making a request, I just looked at the headers on the initialized object. But looks like it adds the header at request time w/ the authorization header

aabmass (approved):

PR #402, opened 2025-04-18
Use uv in instrumentation quickstart

aabmass:
This speeds up the docker build significantly, but also adds a proper locked dependencies for repeatability.


psx95 (approved):
LGTM.

PR #404, opened 2025-05-12
LangGraph console agent with instrumentation sample

aabmass:
This sample has region tags to go in the public docs. It is a CLI based LangGraph agent that is manually instrumented for Cloud Obs. The devsite docs will be recipe focused like
- Add dependencies
- Add instrumentation
- Add manual spans


samples/langgraph-sql-agent/requirements.txt
```diff
@@ -0,0 +1,1289 @@
+# This file was autogenerated by uv via the following command:
+#    uv export -o requirements.txt
```
DylanRussell, line 2: Is this massive file (and the massive uv.lock) expected ?

DylanRussell (approved):

samples/langgraph-sql-agent/requirements.txt
aabmass: Yes uv.lock is autogenerated and this is auto generated from uv.lock

PR #417, opened 2025-07-15
Use `LogRecord.event_name` to set the log ID when the gcp log name attribute is not present.

DylanRussell:


quentinmit:
What is the spec for `event_name`? I wouldn't expect it to be the right grouping for a log name. (A better fit is probably the scope name, for instance.)

DylanRussell:
event name is what makes a log an event : https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-eventname -- why not use it for log name ? We want gen ai events to be easily selected based on the log name drop down in the logs explorer, that's what is motivating this change.. We agreed on this internally, probably better to discuss it there

opentelemetry-exporter-gcp-logging/src/opentelemetry/exporter/cloud_logging/__init__.py
```diff
@@ -197,13 +197,14 @@ def export(self, batch: Sequence[LogData]):
...
-                    attributes.get(
-                        LOG_NAME_ATTRIBUTE_KEY, self.default_log_name
-                    )
+            log_suffix = self.default_log_name
+            log_name_attr = attributes.get(LOG_NAME_ATTRIBUTE_KEY)
+            if log_name_attr and isinstance(log_name_attr, str):
+                log_suffix = urllib.parse.quote_plus(log_name_attr)
+            elif log_record.event_name:
+                log_suffix = urllib.parse.quote_plus(
+                    log_record.event_name.replace(".", "_")
```
aabmass, line 206: Is this needed or the log gets rejected? Wondering if we need it for the LOG_NAME_ATTRIBUTE_KEY case as well

opentelemetry-exporter-gcp-logging/setup.cfg
```diff
@@ -26,8 +26,8 @@ package_dir=
 packages=find_namespace:
 install_requires =
     google-cloud-logging ~= 3.0
-    opentelemetry-sdk ~= 1.30
-    opentelemetry-api ~= 1.30
+    opentelemetry-sdk >= 1.35.0
+    opentelemetry-api >= 1.35.0
```
aabmass, line 30: Was this needed?

opentelemetry-exporter-gcp-logging/src/opentelemetry/exporter/cloud_logging/__init__.py
```diff
@@ -197,13 +197,14 @@ def export(self, batch: Sequence[LogData]):
...
-            log_suffix = urllib.parse.quote_plus(
-                str(
-                    attributes.get(
-                        LOG_NAME_ATTRIBUTE_KEY, self.default_log_name
-                    )
+            log_suffix = self.default_log_name
+            log_name_attr = attributes.get(LOG_NAME_ATTRIBUTE_KEY)
+            if log_name_attr and isinstance(log_name_attr, str):
+                log_suffix = urllib.parse.quote_plus(log_name_attr)
+            elif log_record.event_name:
```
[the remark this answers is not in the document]
DylanRussell: It exists there now -- my PRs moving all the gen ai instrumentations from events to logs use this top level field

opentelemetry-exporter-gcp-logging/src/opentelemetry/exporter/cloud_logging/__init__.py
DylanRussell: actually it's just needed for '/' -> '%2f' conversion... from the proto definition of log entry it says

// `[LOG_ID]` must be less than 512 characters long and can only include the
  // following characters: upper and lower case alphanumeric characters,
  // forward-slash, underscore, hyphen, and period.

Let me update the code to validate this

DylanRussell: I think I had the dot to underscore replacement b/c I thought it looked better, but it isn't needed either really. I'll get rid of that..

opentelemetry-exporter-gcp-logging/src/opentelemetry/exporter/cloud_logging/__init__.py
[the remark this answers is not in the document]
aabmass: opened https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/1078 for the collector

aabmass (approved):

opentelemetry-exporter-gcp-logging/src/opentelemetry/exporter/cloud_logging/__init__.py
```diff
@@ -201,24 +210,28 @@ def __init__(
...
+
     def export(self, batch: Sequence[LogData]):
         now = datetime.datetime.now()
         log_entries = []
         for log_data in batch:
             log_entry = LogEntry()
             log_record = log_data.log_record
             attributes = log_record.attributes or {}
+            if log_record.event_name:
+                attributes["event_name"] = log_record.event_name
```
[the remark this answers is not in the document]
DylanRussell: ok, i switched it to event.name

PR #418, opened 2025-07-15
Add ADK instrumentation sample

psx95:
Sample adopted from https://github.com/aabmass/opentelemetry-operations-python/tree/adk-sample/samples/adk-sql-agent.

### Testing
 - Generates & exports traces to Google Cloud via OTLP.
 - Creates a per session database, multiple sessions work simultaneously each with their isolated copy of sqlite database.
 
 ### TODO
  - Add region tags to support documentation.


samples/adk-sql-agent/main.py
```diff
@@ -0,0 +1,135 @@
...
+    # For this sample, these can be set by loading the `main.env` file.
+    setup_opentelemetry()
+
+    # Call the function to get the FastAPI app instance.
+    # Ensure that the agent director name is the name of directory containing agent subdirectories,
+    # where each subdirectory represents a single agent with __init__.py and agent.py files.
+    # For this example this would be the current directory containing main.py.
+    # Note: Calling this method attempts to set the global tracer provider, which has already been
+    # set by the setup_opentelemetry() function.
+    app: FastAPI = get_fast_api_app(
```
aabmass, line 124: nit I think this was copypasta
```suggestion
    app = get_fast_api_app(
```

samples/adk-sql-agent/sql_agent/agent.py
```diff
@@ -0,0 +1,70 @@
...
+- Always prefer to insert multiple rows in a single call to the sql_db_query tool, if possible.
+- You may request to execute multiple sql_db_query tool calls which will be run in parallel.
+
+You can use run_sql_tool to run sql queries against the current sqlite3 database.
+You can use create_database_tool to create a new ephemeral sqlite3 database if one is not found in current state.
+You should always check if database for current session exist before running sql queries by calling create_database_tool.
+If you make a mistake, try to recover."""
+
+INTRO_TEXT = """\
+Starting agent using ephemeral SQLite DB {dbpath}. This demo allows you to chat with an Agent
```
aabmass, line 44: The dbpath won't get interpolated correctly here. I think we should just rephrase to remove it

samples/adk-sql-agent/sql_agent/agent.py
```diff
@@ -0,0 +1,70 @@
...
+- The sqlite version is {sqlite3.sqlite_version} which supports multiple row inserts.
+- Always prefer to insert multiple rows in a single call to the sql_db_query tool, if possible.
+- You may request to execute multiple sql_db_query tool calls which will be run in parallel.
+
+You can use run_sql_tool to run sql queries against the current sqlite3 database.
+You can use create_database_tool to create a new ephemeral sqlite3 database if one is not found in current state.
+You should always check if database for current session exist before running sql queries by calling create_database_tool.
+If you make a mistake, try to recover."""
+
+INTRO_TEXT = """\
```
aabmass, line 43:
```suggestion
DESCRIPTION = """\
```

samples/adk-sql-agent/sql_agent/tools.py
```diff
@@ -0,0 +1,82 @@
...
+@tracer.start_as_current_span("create_database")
+def create_database_tool(tool_context: ToolContext):
+    """Creates a temporary file in the /tmp directory to hold an ephemeral
+    sqlite3 database if a database is not found for the current session.
+    """
+    if not SESSION_DB_KEY in tool_context.state:
+        _, path = tempfile.mkstemp(suffix=".db")
+        # No scope prefix in the state data indicates that it will be persisted for
+        # current session.
+        # See https://deepwiki.com/google/adk-python/3.4-state-management.
```
aabmass, line 45: Not familiar with DeepWiki, is the URL stable or would it be better to link to https://google.github.io/adk-docs/sessions/state/?

samples/adk-sql-agent/sql_agent/tools.py
```diff
@@ -0,0 +1,82 @@
...
+        tool_context.state[SESSION_DB_KEY] = path
+
+@tracer.start_as_current_span("run_sql")
+def run_sql_tool(sql_query: str, tool_context: ToolContext) -> dict[str, Any]:
+    """Runs a SQLite query. The SQL query can be DDL or DML. Returns the rows if it's a SELECT query."""
+    current_session_db_path = tool_context.state[SESSION_DB_KEY]
```
aabmass, line 51: Should we check if the model hasn't decided to create the session yet and return error ?

samples/adk-sql-agent/sql_agent/tools.py
```diff
@@ -0,0 +1,82 @@
...
+
+class SqlRunResult(TypedDict):
+    error: NotRequired[str]
+    """If the result represents an error. It will be null or undefined if no error"""
+
+    rows: NotRequired[list[tuple[str, ...]]]
+    """The rows returned by the SQL query"""
+
+@tracer.start_as_current_span("create_database")
+def create_database_tool(tool_context: ToolContext):
```
aabmass, line 37: nit add return type

aabmass (approved):
Looks great thank you!

samples/adk-sql-agent/README.md
```diff
@@ -0,0 +1,53 @@
...
+git clone https://github.com/GoogleCloudPlatform/opentelemetry-operations-python.git
+cd opentelemetry-operations-python/samples/adk-sql-agent
+```
+
+Configure the environment and run the sample:
+```sh
+python -m venv venv/
+source venv/bin/activate
+pip install -r requirements.txt
+env $(cat main.env | xargs) python main.py
```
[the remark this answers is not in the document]
psx95: I think that `.env` file in the agent directory holds a special meaning. I could split this `main.env` into 2 files the variables that hold special meaning to the ADK framework can go in `sql_agent/.env` and the the variables that are used to configure OpenTelemetry can go in `opentelemetry.env` -  That way we have a clear separation of concerns.

IMO the environment variable is a better approach since the example is complete even without instructions and the users can see required OTel configuration in one place.

samples/adk-sql-agent/sql_agent/tools.py
psx95: I think it's stable - its copied from adk-python's GitHub repo.
However on second thought your suggestion does seem to be the official docs.

samples/adk-sql-agent/sql_agent/tools.py
psx95: Its a void type, so `def create_database_tool(tool_context: ToolContext) -> None:` ?

samples/adk-sql-agent/sql_agent/tools.py
psx95: We can certainly do that, but, the model should figure out what tool to use and use it accordingly - if the dbpath is null,  then the model should invoke the create_database tool to create one before executing the query.

If we start anticipating exceptions then we should also validate the sql_query before executing it.

psx95: Did some experiments - 
Added a tool that always `raises` exception. It halts the agent's response. It is recommended to always return structured responses from tools even in case of exception.

Based on this, I added a return statement to the create_db tool and added a check. LLM should be able to read the responses and recover.

PR #419, opened 2025-07-15
Drop support for python 3.8

DylanRussell:
Upstream otel python already dropped support for 3.8


aabmass:
You'll have to update the CI checks in branch protection rules to make the "Test 3.8Expected — Waiting for status to be reported" go away

PR #421, opened 2025-07-21
Add region tags & cleanup sample

psx95:
- Add region tags. 
 - Removes unused imports.


aabmass (approved):
Discussed changes offline

PR #432, opened 2025-10-06
Update logic in cloud logging exporter to handle various types in LogRecord.attributes

DylanRussell:


opentelemetry-exporter-gcp-logging/src/opentelemetry/exporter/cloud_logging/__init__.py
```diff
@@ -106,15 +106,27 @@
...
     if isinstance(value, bool):
         return "true" if value else "false"
     if isinstance(value, bytes):
         return base64.b64encode(value).decode()
     if isinstance(value, (int, float, str)):
         return str(value)
-    if isinstance(value, (list, tuple)):
-        return json.dumps(value)
+    if isinstance(value, (list, tuple, Mapping)):
+        return json.dumps(value, separators=(",", ":"), cls=_GenAiJsonEncoder)
```
aabmass, line 124: WDYT of falling back to `str()` to avoid dropping things on the floor? Not sure if this would ever happens because IIRC the sdk does validate the `AnyValue` before it gets enqueued

opentelemetry-exporter-gcp-logging/src/opentelemetry/exporter/cloud_logging/__init__.py
```diff
@@ -176,7 +188,9 @@ def _set_payload_in_log_entry(log_entry: LogEntry, body: AnyValue):
         else:
             log_entry.text_payload = base64.b64encode(body).decode()
     elif body is not None:
-        log_entry.text_payload = _convert_any_value_to_string(body)
+        log_entry.text_payload = _convert_any_value_to_string(
+            body, "LogRecord.body"
```
aabmass, line 192: nit but for `debug_location` i think the more idiomatic thing would be to catch an exception here and log. that way the `code.location` attribute will be correct too

aabmass (approved):

Issue #444, opened 2025-10-31
Remove temporary ReasoningEngine monitored resource support in Cloud Logging exporter


aabmass:
Undo changes in https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/pull/445

PR #445, opened 2025-10-31
Add temporary support for `aiplatform.googleapis.com/ReasoningEngine` monitored resource in logging exporter

aabmass:
Based only on `cloud.resource_id` resource attribute. This gets set in Agent Engine [here](https://github.com/googleapis/python-aiplatform/blob/v1.124.0/vertexai/agent_engines/templates/adk.py#L263-L268).

This may be removed in the future as it's a temporary workaround https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/issues/444

### DIFF
--- a/opentelemetry-exporter-gcp-logging/tests/__snapshots__/test_cloud_logging/test_agent_engine_monitored_resources[client].json
+++ b/opentelemetry-exporter-gcp-logging/tests/__snapshots__/test_cloud_logging/test_agent_engine_monitored_resources[client].json
@@ -6,8 +6,7 @@
         "resource": {
           "labels": {
             "location": "europe-west3",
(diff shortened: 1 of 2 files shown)


opentelemetry-exporter-gcp-logging/tests/__snapshots__/test_cloud_logging/test_agent_engine_monitored_resources[client].json
```diff
@@ -0,0 +1,58 @@
+[
+  {
+    "entries": [
+      {
+        "logName": "projects/fakeproject/logs/test",
+        "resource": {
+          "labels": {
```
Comment (aabmass) [inline opentelemetry-exporter-gcp-logging/tests/__snapshots__/test_cloud_logging/test_agent_engine_monitored_resources[client].json]: Verified the logging API accepts this 

Image

PR #449, opened 2025-11-05
Update sample uv.lock files

aabmass:
Using `uv sync --upgrade`. I verified the samples are working still (and we have test coverage for instrumentation-quickstart). This should fix most of the Dependabot PRs open against this repo.


DylanRussell:
nice

PR #478, opened 2026-03-09
Handle AlreadyExists error in metric descriptor creation

himanhsugusain:
When creating a metric descriptor, catch the AlreadyExists (409) error
and return the local descriptor instead of failing. GCP will update
the descriptor on metric export if needed.

In https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/blob/af5d0725de16c84417f993a21fea3f346e0780c7/opentelemetry-exporter-gcp-monitoring/src/opentelemetry/exporter/cloud_monitoring/__init__.py#L398-L400, if we get metric descriptor already exists exception ( as below ) , the metric emission is skipped.

```
Failed to create metric descriptor labels {
  key: "path"
}
labels {
  key: "opentelemetry_id"
}
....
    raise exceptions.from_grpc_error(exc) from exc
google.api_core.exceptions.AlreadyExists: 409 Errors during metric descriptor creation: {(metric: custom.googleapis.com/app/fastapi/request_count, error: The metric already exists.)}.
```

### DIFF
--- a/opentelemetry-exporter-gcp-monitoring/src/opentelemetry/exporter/cloud_monitoring/__init__.py
+++ b/opentelemetry-exporter-gcp-monitoring/src/opentelemetry/exporter/cloud_monitoring/__init__.py
@@ -20,6 +20,7 @@ from time import time_ns
 from typing import Dict, List, NoReturn, Optional, Set, Union
 
 import google.auth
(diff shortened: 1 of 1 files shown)


himanhsugusain:
GCP return 409 if the metricDescriptor cannot be updated implicitly.

PR #540, opened 2026-07-03
Fix typo s/Exporting more then/Exporting more than/g

toransahu:


DylanRussell:
We are going to deprecate this exporter and point people to the python OTLP exporters

PR #550, opened 2026-07-29
Release pr/1.13.0 for exporters

DylanRussell:


aabmass:
Nothing in the changelogs, what do we need to release here?

DylanRussell:
Forgot to add the changelog in my deprecation PR.. I've added it into this PR now.... PTAL

aabmass:
Thanks. One other question, how did you disable the resource detector from the release?

DylanRussell:
I had jetski temporarily add a --package_glob option to release.py and ran it with:  --package_glob "opentelemetry-exporter-*"

aabmass:
I think we should commit it in for later release, or just delete the code, and add something [like this](https://pypi.org/classifiers/#:~:text=To%20prevent%20a%20package%20from%20being%20uploaded%20to%20PyPI%2C%20use%20the%20special%20%22Private%20%3A%3A%20Do%20Not%20Upload%22%20classifier.%20PyPI%20will%20always%20reject%20packages%20with%20classifiers%20beginning%20with%20%22Private%20%3A%3A%22.) to prevent accidental uploading from this repo. I haven't removed our PyPI account from the package, so nothing stops accidentally uploading to the package.

DylanRussell:
Ok lets just delete it.. I've opened up https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/pull/551 -- that docs directory is not something that's part of the PyPi release of the exporters right ?

aabmass (approved):

PR #559, opened 2026-07-30
Add force_flush method to CloudLoggingExporter

DylanRussell:
v1.44 of otel sdk breaks the exporter because it requires a force_flush method on logging exporters... This PR fixes it


opentelemetry-exporter-gcp-logging/src/opentelemetry/exporter/cloud_logging/__init__.py
```diff
@@ -474,5 +476,13 @@ def _write_log_entries_to_client(
                     "Error while writing to Cloud Logging", exc_info=ex
                 )
 
+    def force_flush(self, timeout_millis: float = 30_000) -> bool:
+        """Flushes any buffered logs.
+
+        For CloudLoggingExporter, this is a no-op as logs are exported synchronously.
+        """
+        logger.info("force_flush does nothing for CloudLoggingExporter")
```
aabmass: not sure it's needed or maybe debug level

aabmass (approved):

